PR #1, opened 2017-10-26
lint fixes

dags:
I swiched back to 'app" prefix, instead of former 'fc' one.


src/assets/demo/workspaces.json
```diff
@@ -89,7 +89,7 @@
             "groupName": "ee910c59-09cb-48a7-9e04-8380787b677f-WRITER"
           }
         },
-        "bucketName": "fc-ee910c59-09cb-48a7-9e04-8380787b677f",
+        "bucketName": "app-ee910c59-09cb-48a7-9e04-8380787b677f",
```
pshapiro4broad: Did you mean to change `fc` in this file as well? This is test data (and it's currently unused) so the change will have no effect, but it seems like it wasn't an intentional change.

dags:
Lint changes will go in a newer PR.

PR #4, opened 2017-11-07
DSDEGP-1681 - Authenticate to FireCloud and Login

LeoForconesi:
### Description
_* JIRA related story: https://broadinstitute.atlassian.net/browse/DSDEGP-1681_
_* Creation of a login service and component to handle google authentication to Fire Cloud Explorer App_
_* Get authorization token to authenticate all API calls to Fire Cloud and Google Storage_
_* Integration of ngx-electron library to wrap all electron's functionality into a typescript service_
_* Re-directions from login to home and vice-versa_

----

### Checklist

* Review cycle:
- [x] **Submitter**: Move the JIRA issue to QA once this checklist is completed


pshapiro4broad:
Can you run `npm run lint` locally and fix the lint errors? Once we have the CI build working these errors will be more obvious.

src/app/app-routing.module.ts
```diff
@@ -1,20 +1,24 @@
...
+    redirectTo: 'login',
     pathMatch: 'full'
   },
     {
     path: 'home',
     component: HomeComponent
+    },
+  {
+    path: 'login',
+    component: LoginComponent
```
pshapiro4broad: It's not clear to me how `LoginComponent` is used here. It looks like login works via `AppComponent.onLoginClick()` which is called from the `Sign In` button. I tried removing `LoginComponent` and its associated routes and it didn't seem to affect the login code behavior.

src/app/file-manager/services/login.service.ts
```diff
@@ -0,0 +1,54 @@
...
+const googleOptions = {
+  scope: ['email',
+          'profile',
+          'https://www.googleapis.com/auth/compute',
+          'https://www.googleapis.com/auth/devstorage.full_control'].join(' '),
+  accessType: 'offline'
+};
+
+@Injectable()
+export class LoginService {
```
pshapiro4broad: We will need a mocked version of this service for testing. You can do this now or it can wait until the a test needs it.

src/app/app.component.ts
```diff
@@ -59,11 +66,20 @@ export class AppComponent implements OnInit, OnDestroy {
...
+      this.loginService.googleLogin().then(
+          res => this.router.navigate([redirect]));
+    } else {
+      this.router.navigate([redirect]);
+    }
   }
 
   onLogoutClick() {
+    const redirect = '/login';
+    SecurityService.removeAccessToken();
```
pshapiro4broad: Even after clicking `Log Out` I don't get prompted for username/password when I click Log In. I also noticed that my login token was saved between runs. Is this the intended behavior? It seems like we should have a way to revoke the token and force the user to re-enter their credentials. If this is working as intended, we can talk about this with the stakeholders; maybe this behavior would be preferred.

src/app/app-routing.module.ts
LeoForconesi: The idea of `LoginComponent` is basically to deliver a previous step to access any work space download list without showing an empty list. Here we can notify if a user whether is allowed to access or it should register to Firecloud in order to access our app. 

At the moment this is empty since is to be implemented, but i can remove it from this branch if you consider its unnecessary right now.

src/app/file-manager/services/login.service.ts
LeoForconesi: If it's ok for you, we could do this on the next sprint so we can integrate this important functionality on develop, and we could create a story on Jira to track it down.

src/app/app-routing.module.ts
pshapiro4broad: I think it's better to wait until it's needed to add it. You can make a story for the requirement you're thinking of and attach the code to the JIRA if you want to make sure the work isn't lost.

src/app/file-manager/services/login.service.ts
pshapiro4broad: Sure, that would be fine.

src/app/app.component.ts
LeoForconesi: Yes, the login token is saved when a usr logs in, and is flushed when logged out. At the moment it would be the intended behavior, but we need to perform a log out from google's account besides from our app (which are two different things). Its ok if we create a story to handle this more clearly?

pshapiro4broad: Yes, that's OK

PR #7, opened 2017-11-13
UI Refinements - Lint errors fix

LeoForconesi:
### Description

_**Jira's Story:** https://broadinstitute.atlassian.net/browse/DSDEGP-1738_
_UI refinements for first iteration demo._
_Code-cleaning to make the demo look more like the wireframe reviewed with the stakeholders._
_Fix for Lint errors_

----

### Checklist

* Review cycle:

### DIFF
--- a/src/app/file-manager/file-explorer/file-explorer.component.ts
+++ b/src/app/file-manager/file-explorer/file-explorer.component.ts
@@ -24,13 +24,16 @@ export class FileExplorerComponent implements OnInit {
   msgs: Message[];
 
   @Output('done') done: EventEmitter<any> = new EventEmitter();
+  @Output('count') count: EventEmitter<any> = new EventEmitter();
 
   files: TreeNode[];
   workspaces: any[];
 
-  selectedFiles: TreeNode[];
+  selectedCount = 0;
+  selectedFiles: TreeNode[] = [];
   selectedFile: TreeNode;
 
+  folderCount = 0;
   fileCount = 0;
   fileSize = 0;
 
@@ -59,7 +62,7 @@ export class FileExplorerComponent implements OnInit {
     this.cols = [
       { field: 'path', header: 'Name', footer: 'Name' },
       { field: 'size', header: 'Size', footer: 'Size' },
-      { field: 'type', header: 'Type', footer: 'Type' }
+      { field: 'updated', header: 'Modified', footer: 'Modified' }
     ];
 
     this.items = [
@@ -68,12 +71,23 @@ export class FileExplorerComponent implements OnInit {
     ];
   }
 
+  countFiles() {
+    this.fileCount = 0;
+    this.selectedFiles.forEach(f => {
+      if (f.data.type === 'File') {
+        this.fileCount++;
+      }
+    });
+  }
+
   nodeSelect(event) {
+    this.countFiles();
     this.msgs = [];
     this.msgs.push({ severity: 'info', summary: 'Node Selected', detail: event.node.data.name });
   }
 
   nodeUnselect(event) {
+    this.countFiles();
     this.msgs = [];
     this.msgs.push({ severity: 'info', summary: 'Node Unselected', detail: event.node.data.name });
   }
@@ -90,6 +104,7 @@ export class FileExplorerComponent implements OnInit {
 
   deleteNode(node: TreeNode) {
     node.parent.children = node.parent.children.filter(n => n.data !== node.data);
+    this.countFiles();
     this.msgs = [];
     this.msgs.push({ severity: 'info', summary: 'Node Deleted', detail: node.data.name });
   }
@@ -120,10 +135,12 @@ export class FileExplorerComponent implements OnInit {
     this.files.forEach(node => {
       this.selectRecursive(node, true);
     });
+    this.countFiles();
   }
 
   selectNone() {
     this.selectedFiles = [];
+    this.countFiles();
   }
 
   toggleSelection() {
@@ -136,6 +153,7 @@ export class FileExplorerComponent implements OnInit {
       }
     });
     this.selectedFiles = newSelection;
+    this.countFiles();
   }
 
   private selectRecursive(node: TreeNode, isExpand: boolean) {
--- a/src/app/file-manager/reducers/downloadables.reducer.ts
+++ b/src/app/file-manager/reducers/downloadables.reducer.ts
@@ -61,6 +61,21 @@ export function DownloadablesReducer(state = initialState, action: Action): Down
                 })
             };
 
+        case DownloadablesActions.TOGGLE_ITEM_SELECTION:
+
+            const tmp_tog = state.items.filter(item => {
+                if (item.id === action.payload.id) {
+                    item.selected = !item.selected;
+                }
+                return true;
+            });
+
+            return {
+                count: tmp_tog.length,
+                selectedCount: tmp_tog.filter(x => x.selected === true).length,
+                items: tmp_tog
+            };
+
         case DownloadablesActions.ADD_ITEM:
             const add_item = [...state.items, action.payload];
             const add_item_sc = add_item.filter(x => x.selected === true).length;
@@ -97,14 +112,10 @@ export function DownloadablesReducer(state = initialState, action: Action): Down
                 }
                 return true;
             });
-            const sel_item_sc = sel_item.filter(item => {
-                return item.selected === true;
-            });
 
-            const sel_item_count = state.count;
             return {
-                count: sel_item_count,
-                selectedCount: sel_item_sc.length,
+                count: state.count,
+                selectedCount: sel_item.filter(item => item.selected === true).length,
                 items: sel_item
             };
 
--- a/src/app/file-manager/transferables-grid/transferables-grid.component.ts
+++ b/src/app/file-manager/transferables-grid/transferables-grid.component.ts
@@ -1,46 +1,139 @@
-import { Component, OnInit, ViewChild, ChangeDetectionStrategy } from '@angular/core';
+import { Component, OnInit, AfterViewInit, Input, ViewChild, ChangeDetectionStrategy } from '@angular/core';
 
 import { Store } from '@ngrx/store';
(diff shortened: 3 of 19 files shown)


src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -24,13 +24,16 @@ export class FileExplorerComponent implements OnInit {
...
   workspaces: any[];
 
-  selectedFiles: TreeNode[];
+  selectedCount = 0;
+  selectedFiles: TreeNode[] = [];
   selectedFile: TreeNode;
 
+  folderCount = 0;
   fileCount = 0;
   fileSize = 0;
```
pshapiro4broad: This code has many unused variables, such as `workspaces`, `selectedCount`, `folderCount`, `fileSize`. The constructor accepts `gcsService` and `firecloudService` but they are both unused as well. Unused variables make the code harder to read and maintain, so they should be deleted if they're not used.

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -24,13 +24,16 @@ export class FileExplorerComponent implements OnInit {
   msgs: Message[];
 
   @Output('done') done: EventEmitter<any> = new EventEmitter();
```
pshapiro4broad: Is `'done'` connected to any `@Input`? I don't see a reference to it from another component.

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -24,13 +24,16 @@ export class FileExplorerComponent implements OnInit {
...
+  @Output('count') count: EventEmitter<any> = new EventEmitter();
```
pshapiro4broad: I see that `'count`` is used by `TransferablesGridComponent` but it doesn't appear to be ever set with a value here. If it's not being used it should be removed.

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -59,7 +62,7 @@ export class FileExplorerComponent implements OnInit {
     this.cols = [
       { field: 'path', header: 'Name', footer: 'Name' },
       { field: 'size', header: 'Size', footer: 'Size' },
-      { field: 'type', header: 'Type', footer: 'Type' }
+      { field: 'updated', header: 'Modified', footer: 'Modified' }
     ];
 
     this.items = [
```
pshapiro4broad: These menu items are bound and have commands but I don't think we need View or Delete for this app. Also I wasn't able to see them by bringing up the context menu; does `p-contextMenu` work in electron?

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -68,12 +71,23 @@ export class FileExplorerComponent implements OnInit {
     ];
   }
 
+  countFiles() {
+    this.fileCount = 0;
+    this.selectedFiles.forEach(f => {
```
pshapiro4broad: Is there a way to observe `selectedFiles` in angular? If so, you could replace the explicit method calls to `countFIles()` with an observer/subscribe pattern.

src/app/file-manager/reducers/downloadables.reducer.ts
```diff
@@ -61,6 +61,21 @@ export function DownloadablesReducer(state = initialState, action: Action): Down
...
+
+            const tmp_tog = state.items.filter(item => {
+                if (item.id === action.payload.id) {
+                    item.selected = !item.selected;
+                }
+                return true;
+            });
+
+            return {
+                count: tmp_tog.length,
```
pshapiro4broad: These values are computed in the same way for each `return` here. It would be clearer and more concise to use a constructor for `DownloadableState` that accepts `Item[]`, since the other elements `DownloadableState` can be derived from the array argument.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
```diff
@@ -1,54 +1,147 @@
...
+  getSortedData(data: Item[]): Item[] {
+
+    if (!this._sort.active || this._sort.direction === '') { return data; }
+
+    return data.sort((a, b) => {
+      let propertyA: number | string = '';
+      let propertyB: number | string = '';
+      const propertyC: Date | string = '';
+
+      switch (this._sort.active) {
```
pshapiro4broad: Instead of doing the `switch` inside the `sort()`, you could `switch()` first to generate the sort method and then pass the correct method to the `sort`, to avoid running `switch` every time you compare two items.

src/app/file-manager/file-explorer/file-explorer.component.ts
LeoForconesi: Yes, it is used in home.component.html component to change the tab.

src/app/file-manager/file-explorer/file-explorer.component.ts
LeoForconesi: it is used to pass the selected count value from file-explorer.component.ts to transferable-grid.component.ts. It is set on home.component.html.

pshapiro4broad: I'm still learning angular so I'm probably missing something here, but I see `<app-file-explorer ... (count)="track($event)"` in `home.component.html`. My understanding was that this binds `count` as an output of `app-file-explorer`. But `count` (unlike `done`) doesn't appear to be set anywhere in `FileExplorerComponent` so its value will always be `0`. I tried setting a breakpoint on `HomeComponent.track()` and wasn't able to make it trigger.

src/app/file-manager/reducers/downloadables.reducer.ts
LeoForconesi: DownloadableState is an interface and there is no actual class implementing it. That is why we use it that way in the reducer.  If we really have to use that constructor we'll need to add the class.

pshapiro4broad: From my understanding of typescript you can replace the interface with a class of the same name without changing anything else. E.g.
```
export class DownloadableState {
    count: number;
    selectedCount: number;
    items: Item[];
    constructor(items: Item[]) {
      this.items = items;
      this.count = items.length;
      this.selectedCount = items.filter(x => x.selected === true).length;
    }
}
```

and this code becomes
```
return new DownloadableState(tmp_tog);
```

LeoForconesi: yes, we just did it that way. Thanks.

src/app/file-manager/file-explorer/file-explorer.component.ts
LeoForconesi: Since we don't modify selectedFiles directly, it is modified by Primeng's p-treeTable ([(selection)]="selectedFiles") it is not that easy to implement the Observable approach. I suggest to postpone this until all pending PR are merged.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
LeoForconesi: We are not actually using sort, we'll remove it.

src/app/file-manager/reducers/downloadables.reducer.ts
```diff
@@ -61,6 +61,21 @@ export function DownloadablesReducer(state = initialState, action: Action): Down
                 })
             };
 
+        case DownloadablesActions.TOGGLE_ITEM_SELECTION:
+
+            const tmp_tog = state.items.filter(item => {
```
pshapiro4broad: this seems like an odd way to modify a list of items, normally I'd use `map()` for this, e.g.
```
const tmp_tog = state.items.map(item => item.selected = !item.selected);
```
Although this modifies `state.items` in place so there's no need to declare `tmp_tog`, using `state.items` should have the same effect.

src/app/file-manager/file-explorer/file-explorer.component.ts
LeoForconesi: Unused variables have been removed

src/app/file-manager/file-explorer/file-explorer.component.ts
LeoForconesi: Context menu has been removed.

src/app/file-manager/file-explorer/file-explorer.component.ts
dags: The emitter was never called, but now ii is removed and replaced with datasource.data.length number. Cleaner, shorter code, I guess.

src/app/file-manager/reducers/downloadables.reducer.ts
```diff
@@ -11,6 +11,26 @@ export interface DownloadableState {
     items: Item[];
 }
 
+export class DownloadableStateImpl implements DownloadableState {
+    itms: Item[];
```
pshapiro4broad: It would be clearer to not have a getter for `items` and instead keep a field called `items`. In the case where we do have a name conflict, it looks like the convention is to name the field using an underscore, e.g. `_items`.

src/app/file-manager/reducers/downloadables.reducer.ts
```diff
@@ -11,6 +11,26 @@ export interface DownloadableState {
...
+
+    constructor(itm: Item[]) {
```
pshapiro4broad: There is no need to use a different spelling here, as members are always prefixed by `this.` so there is no ambiguity. E.g.
```
constructor(items: Item[]) {
  this.items = items;
}
```

src/app/file-manager/reducers/downloadables.reducer.ts
```diff
@@ -11,6 +11,26 @@ export interface DownloadableState {
     items: Item[];
 }
 
+export class DownloadableStateImpl implements DownloadableState {
```
pshapiro4broad: This is OK but I don't see why you didn't change `DownloadableState` to be a concrete class instead of adding an Impl class. As far as I can tell that would work fine and would avoid adding another class type to the code.

PR #8, opened 2017-11-15
DSDEGP-1710: get automated tests working

dags:
### Description
Updated test to avoid failures while improving Travis build.
----

### Checklist
Dsdegp 1710


src/app/app.component.spec.ts
```diff
@@ -1,11 +1,37 @@
...
+import { NgModule, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
+import { SharedModule } from '@app/shared';
+import { CoreModule } from '@app/core';
+import { FileManagerModule } from '@app/file-manager';
+import { SettingsModule } from './settings';
+import { NgxElectronModule } from 'ngx-electron';
+import { RouterTestingModule } from '@angular/router/testing';
+
 describe('AppComponent', () => {
   beforeEach(async(() => {
```
pshapiro4broad: Is `async()` needed here, and below in the `it()` clauses? Other examples I've seen don't use this construct.

dgil-olx: Spec files are created by default by Angular CLI tool so I think follows Angular best practices. I need to do some research before being able to confirm it is a bad practice. Could we put this on testing technical debt ?

pshapiro4broad: Yes that would be OK

dags: Spec files are created by default by Angular CLI tool so I think follows Angular best practices. I need to do some research before being able to confirm it is a bad practice. Could we put this on testing technical debt ?

PR #10, opened 2017-11-16
DSDEGP-1739 - Workbench files list page improvements

vvicario:
### Description

_Please explain the changes you made here._
_Explain the **motivation** for making this change. What existing problem does the pull request solve?_
_Mention any JIRA issues fixed, addressed or otherwise related to this pull request, including issue numbers or hard links for issues in other repos._

https://broadinstitute.atlassian.net/browse/DSDEGP-1739

----

### Checklist

* Review cycle:


src/app/file-manager/filters/filesize-filter.spec.ts
```diff
@@ -0,0 +1,37 @@
+import { TestBed, inject } from '@angular/core/testing';
+
+import { FilterSizePipe } from './filesize-filter';
+
+describe('SecurityService', () => {
```
pshapiro4broad: test name should be `FilterSizePipe`

src/app/file-manager/filters/filesize-filter.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      providers: [FilterSizePipe]
+    });
+  });
+  const BYTES = ' Bytes';
+  const KB = ' KB';
+  const MB = ' MB';
+  const GB = ' GB';
+  const TB = ' TB';
```
pshapiro4broad: It would be better to define these constants in `FilterSizePipe` so you can reference them in the test code and wouldn't need to duplicate them here.

src/app/file-manager/services/files.service.ts
```diff
@@ -95,47 +97,42 @@ export class FilesService {
...
         treeLocal.push(node);
         return treeLocal;
     }
 
     private createNode(item: any, path: string): TreeNode {
         const node: TreeNode = {};
+        const fileData: FileData = this.getFileData(item, path);
+
+        node.data = fileData ;
+        return node;
```
pshapiro4broad: you could return the tree node here directly, e.g.
```
  private createNode(item: any, path: string): TreeNode {
    return {data: this.getFileData(item, path)};
  }
```

src/app/file-manager/services/files.service.ts
```diff
@@ -95,47 +97,42 @@ export class FilesService {
...
+    private getFileData(item: any, path: string) {
         const fileData: FileData = {
             id: item.id,
             selfLink: item.selfLink,
             bucket: item.bucket,
             created: item.timeCreated,
             updated: item.updated,
             path: path,
-            size: item.size + ' MB',
+            size: this.filterSize.transform(item.size),
```
pshapiro4broad: This works but I thought the angular way to do this was to use a custom filter, and use a filter in the UI, not modify the data.

vvicario: Given the way we show file data using a primeng's p-treeTable : 

<p-treeTable [value]="files" selectionMode="checkbox" [(selection)]="selectedFiles" styleClass="tree-table" (onNodeSelect)="nodeSelect($event)"
  (onNodeUnselect)="nodeUnselect($event)">
  *<p-column *ngFor="let col of cols " [field]="col.field" [header]="col.header"></p-column>*
</p-treeTable>
we should explicitly create one p-column for each field to be able to use pipes or other UI formatting to the size field. Currently, we can modify which columns are shown just changing cols array. 

We found it is more appropriate to modify data once and use it N times instead of transforming data on each place where it is shown.

If you still prefer to have pipes / formatting at column level instead of data array, please let me know and I will change it asap.

pshapiro4broad: OK, I understand. I'm more used to seeing code where the data exported to the UI isn't formatted or filtered and the formatting is done in the HTML code. It's OK to leave it this way for now.

PR #12, opened 2017-11-21
DSDEGP-1746 UI prompts the user for the location on the local file system where the files should be copied 

vvicario:
### Description

Download File: Let the user select the folder where the data should be copied.

----

### Checklist

- [X ] **Submitter**: Ensure that you have added or updated tests
* Review cycle:


src/app/file-manager/actions/downloadables.actions.ts
```diff
@@ -66,6 +68,19 @@ export class SelectItem implements Action {
     constructor(public payload: any) { }
 }
 
+export class SetDestination implements Action {
+    readonly type = SET_DESTINATION;
+
+    constructor(public payload: any) { }
```
pshapiro4broad: is it possible to use a more narrow type here, e.g. `string` for this payload and `boolean` for the other?

src/app/file-manager/file-manager.module.ts
```diff
@@ -28,6 +26,8 @@ import { HTTP_INTERCEPTORS } from '@angular/common/http';
 import { RequestInterceptor } from './services/request.interceptor';
 import { environment } from '@env/environment';
 import { HttpClient } from '@angular/common/http';
+import { FileModalComponent } from './file-modal/file-modal.component';
```
pshapiro4broad: Is there a reason why the import paths use `'./file-modal/` instead of `file-modal/`? It's OK either way but I wasn't sure if the usage was intentional or not.

src/app/file-manager/file-modal/file-modal.component.html
```diff
@@ -0,0 +1,19 @@
...
+  <div class="flex-container">
+      <div class="browse-button">
+          <label for="directorypicker">Browse...</label>
+          <input type="file" id="directorypicker"  (change)="directoryChanged($event)" style="display: none;" webkitdirectory directory multiple required/>
+      </div>
+    <p class="browse-location">{{directory}}</p>
+    
+  </div>
+  <hr>
+  <mat-checkbox aria-label="Preserve folder structure"  (change)="selectionChanged($event)" color="primary" class="dialog-checkbox" disableRipple>Preserve local structure</mat-checkbox>
```
pshapiro4broad: I think the label 'Preserve local structure' is confusing as the structure being preserved is the remote (cloud) structure not the local one. I think 'Preserve folder structure' is better for the checkbox label.

src/app/file-manager/file-modal/file-modal.component.html
```diff
@@ -0,0 +1,19 @@
+<h2 mat-dialog-title>Where do you want to download to?</h2>
+
+<mat-dialog-content>
+  <div class="flex-container">
+      <div class="browse-button">
+          <label for="directorypicker">Browse...</label>
+          <input type="file" id="directorypicker"  (change)="directoryChanged($event)" style="display: none;" webkitdirectory directory multiple required/>
```
pshapiro4broad: When I tried this out I noticed that the Browse button worked but didn't have a visible response to a mouse click. Is there a way to fix that?

src/app/file-manager/transferables-grid/transferables-grid.component.ts
```diff
@@ -191,4 +204,12 @@ export class TransferablesGridComponent implements OnInit, AfterViewInit {
 
   sortData(evt) {
   }
+
+  startDownload() {
+    /* Info needed to download file:
+    console.log(JSON.stringify(this.filesDatabase.data));
```
pshapiro4broad: Does the `data` field include the workspace root? We'll need both the path to the file and the workspace root to determine the relative path that needs to be created on the local disk.

pshapiro4broad:
I ran this and clicked around a bit and got a few errors in the console:
```
ng:///FileManagerModule/FileModalComponent.ngfactory.js:22 ERROR TypeError: Cannot read property 'path' of undefined
    at FileModalComponent.webpackJsonp.../../../../../src/app/file-manager/file-modal/file-modal.component.ts.FileModalComponent.directoryChanged (main.bundle.js:1176)
    at Object.eval [as handleEvent] (ng:///FileManagerModule/FileModalComponent.ngfactory.js:28)
    at handleEvent (vendor.bundle.js:115787)
    at callWithDebugContext (vendor.bundle.js:117272)
    at Object.debugHandleEvent [as handleEvent] (vendor.bundle.js:116859)
    at dispatchEvent (vendor.bundle.js:112236)
    at vendor.bundle.js:112850
    at HTMLInputElement.<anonymous> (vendor.bundle.js:159559)
    at ZoneDelegate.webpackJsonp.../../../../zone.js/dist/zone.js.ZoneDelegate.invokeTask (polyfills.bundle.js:6565)
    at Object.onInvokeTask (vendor.bundle.js:107150)
```

```
ng:///MatDialogModule/MatDialogContainer_Host.ngfactory.js:5 ERROR TypeError: Cannot read property 'directory' of undefined
    at SafeSubscriber._next (main.bundle.js:898)
    at SafeSubscriber.webpackJsonp.../../../../rxjs/Subscriber.js.SafeSubscriber.__tryOrUnsub (vendor.bundle.js:24534)
    at SafeSubscriber.webpackJsonp.../../../../rxjs/Subscriber.js.SafeSubscriber.next (vendor.bundle.js:24481)
    at Subscriber.webpackJsonp.../../../../rxjs/Subscriber.js.Subscriber._next (vendor.bundle.js:24421)
    at Subscriber.webpackJsonp.../../../../rxjs/Subscriber.js.Subscriber.next (vendor.bundle.js:24385)
    at Subject.webpackJsonp.../../../../rxjs/Subject.js.Subject.next (vendor.bundle.js:24129)
    at SafeSubscriber._next (vendor.bundle.js:138832)
    at SafeSubscriber.webpackJsonp.../../../../rxjs/Subscriber.js.SafeSubscriber.__tryOrUnsub (vendor.bundle.js:24534)
    at SafeSubscriber.webpackJsonp.../../../../rxjs/Subscriber.js.SafeSubscriber.next (vendor.bundle.js:24481)
    at Subscriber.webpackJsonp.../../../../rxjs/Subscriber.js.Subscriber._next (vendor.bundle.js:24421)
```

src/app/file-manager/transferables-grid/transferables-grid.component.ts
tadeork: I think there's no need to have the workspace in the `data` because it's already in the info that I'm getting from the selected files to download in the `file-explorer.component.ts`

pshapiro4broad: Maybe there's a problem with the way state is being stored right now. Right now `TransferablesGridComponent` has a single object `FilesDatabase` which contains `destination` and `preserveStructure`. But there could be different downloads running at the same time that have different settings for these objects. So this state should be stored in the model that backs `TransferableItemComponent` instead.

I think we need a model that contains items that are downloading (or are queued/paused in a download). Each downloading object contains its source path and its destination path; the original values for `destination` and `preserveStructure` would no longer be necessary at this point. It looks like `FilesDatabase` is this model, but the model doesn't have all the data we need yet.

It's important that the transferable grid be able to contain more than one set of selected downloads at a time, each of which could have a different set of download settings.

It also occurs to me that each downloadable item may have its own download controller, because in the case where we allow upload, download and copy between different services, we'd need to make the pause/resume/abort operations behave differently based on the item type. This controller (or service) would also be set at the time the downloading object is first created.

src/app/file-manager/file-manager.module.ts
vvicario: Yes, it was intentional because the module specifier need to start with './'. Basically if you're wanting to reference a relative module you must start the module identifier with a dot, so either './' or '../'. Otherwise the module is considered to be a "global" module

src/app/file-manager/file-modal/file-modal.component.html
vvicario: can you please give us more details?

pshapiro4broad: Sure, the behavior I see is when I click on the Browse button, it doesn't change its visible state. I would expect it to become darker, like the Cancel button does when it's clicked on.

image

In this screen shot the mouse is clicking on the Cancel button, so it's gray, not white. When I click on the Browse... button it stays white. This has the effect of making it look like the button isn't being clicked on.

vvicario:
@pshapiro4broad  code is ready, please let me know if you have any other feedback

pshapiro4broad:
Two more things I noticed when testing:

- the checkbox next to 'Preserve folder structure' is sometimes missing; I'm using chrome on MacOS if that makes a difference. I saw the checkbox the first time, but when I go back and start another download it doesn't appear, although it does flash onscreen when I click on the checkbox label text

- the directory setting is not preserved; when the user brings up the dialog again the directory field is blank. If this isn't part of this checkin, create another JIRA ticket to cover this feature

pshapiro4broad:
Not sure if it helps, but here's the inspector info for the checkbox when hidden

missing checkbox

wloforte:
@pshapiro4broad We couldn't reproduce this issue easily in our local environment but I think I was able to fix it. If you can reproduce this again please let us know and I'll create a ticket (bug) so we can fix it.

vvicario:
@pshapiro4broad  the directory setting is  preserved now, we will merge so we can continue with US 1761, if you have any other comment, we will create a new branch and work on that.

PR #13, opened 2017-11-22
DSDEGP-1768 - Verify if User is registered to FC at login

LeoForconesi:
### Description

_This code checks if the user is registered to Firecloud site at the login, providing some messages describing possible solutions. Eg. to register to firecloud._
_The motivation for this code is to provide an extra security level to users that uses this app and provide recovery options from possible system errors or user mistakes at the login_
_Jira's TIcket link: https://broadinstitute.atlassian.net/browse/DSDEGP-1768_

----

### Checklist

* Review cycle:

### DIFF
--- a/src/app/file-manager/services/login.service.ts
+++ b/src/app/file-manager/services/login.service.ts
@@ -1,9 +1,11 @@
 import { Injectable } from '@angular/core';
 import { environment } from '@env/environment';
 import { Store } from '@ngrx/store';
-import { login} from '@app/core';
+import { login } from '@app/core';
 import { ElectronService } from 'ngx-electron';
 import { SecurityService } from '../services/security.service';
+import { FirecloudService } from '../services/firecloud.service';
+import { Observable } from 'rxjs/Observable';
 
 const googleConfig = {
   clientId: environment.CLIENT_ID,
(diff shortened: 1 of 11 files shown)


pshapiro4broad:
One problem I can see while testing this is that, in the case where the cookie is used to log in, the Sign In page isn't locked out, and no UI feedback is presented that the user should wait. So if the user clicks Sign In, sees the google dialog come up and go away quickly, and is returned to the Sign In page, they can click the button again. Instead we want a UI flow where they're locked out of this. One way is to eliminate our Sign In page, e.g. the user is always sent to the google login dialog right away. We would have to clear the google cookies on Sign Out, so it has an effect.

src/app/file-manager/services/login.service.ts
```diff
@@ -16,34 +18,52 @@ const googleConfig = {
...
+      }
+      );
     } else {
       alert('Implement web login');
       return null;
     }
   }
 
+  setGoogleOptions(): Observable<any> {
+    return new Observable<any>(
```
pshapiro4broad: Do you need to make a new observable each time this method is called? Or could we make this object once and return the same one each time?

LeoForconesi: This observable has been removed and replaced by a constructor call to configure google options.

LeoForconesi:
You're right, this a flaw we need to correct.
Regarding google's dialog, instead of cleaning the app's cache it would be preferable to perform a proper google's account logout. For this we got an existing Jira's story that we could implement on the upcoming sprint if its ok for you. 

On the other hand, to disable or block electron's parent window when google's pop-up is displayed we need to investigate how to accomplish that. Maybe we could add this aspect as an acceptance criteria in google's logout story. https://broadinstitute.atlassian.net/browse/DSDEGP-1776
I'll wait for your comments on how to proceed with this.

LeoForconesi:
Update. 
- We managed a workaround to lock Sign in page.
- We would need to define and implement the need of a Ui feedback such as a load spinner for loading moments.

PR #16, opened 2017-11-29
DSDEGP-1761 Implement Download Process for a Single File

tadeork:
### Description

_Implemented single file download. Creation of folder structure if required._
_In this storie we need to have a download process for a single file at a time and, if the user requires, reproduce the folder structure as is in FC Service._
_JIRA [single download story](https://broadinstitute.atlassian.net/browse/DSDEGP-1761), [electron app folders story](https://broadinstitute.atlassian.net/browse/DSDEGP-1791)_

----

### Checklist

* Review cycle:


src/app/file-manager/services/register-download.service.ts
```diff
@@ -0,0 +1,16 @@
...
+import { ElectronService } from 'ngx-electron';
+import { SecurityService } from './security.service';
+import { Item } from '../models/item';
+
+@Injectable()
+export class RegisterDownloadService {
+
+  constructor(private electronService: ElectronService) { }
+
+  public _downloadEvent(items: Item[]) {
```
pshapiro4broad: Why is this called `_downloadEvent()` instead of `downloadEvent()`? Does it conflict with another name? Also it seems like a better name for this is `startDownload()`

main.js
```diff
@@ -74,6 +75,9 @@ app.on('ready', function () {
   });
   // ----- Google auth -----
 
+  ipcMain.on('start-download', (event, items, access_token) => {
```
pshapiro4broad: Can we define constants for these IPC message names to avoid magic strings in the code?

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -26,6 +27,8 @@ export class FileExplorerComponent implements OnInit {
   @Output('done') done: EventEmitter<any> = new EventEmitter();
 
   files: TreeNode[];
+  item: Item;
```
pshapiro4broad: At this point I think we need comments for each field in the component to describe what state it represents. `files` seems pretty clear, it's the model that this component is drawing onscreen. `selectedFiles` is also pretty clear, it's the state of the current selection. But things like `item` and `downloadItems` and `archivos` are less clear to me.

src/app/file-manager/services/register-download.service.ts
tadeork: No, it's just a convention in some libraries that I follow. I'll change it.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
```diff
@@ -195,7 +196,6 @@ export class TransferablesGridComponent implements OnInit, AfterViewInit {
   }
 
   startDownload() {
```
pshapiro4broad: Why is this operation here and not in `FileExplorerComponent`?

src/app/file-manager/file-explorer/file-explorer.component.ts
tadeork: Sorry, you're totally right. I'll forget to change those names.

electron_app/helpers/_handleDisk.js
```diff
@@ -0,0 +1,27 @@
+const fs = require('fs');
```
pshapiro4broad: Is it typical for helper files to start with an underscore?

Also the file is `_handleDisk.js` but the method it contains is `handleFolder`, shouldn't these names be the same?

electron_app/helpers/_handleDisk.js
```diff
@@ -0,0 +1,27 @@
...
+        };
+    return new Promise((resolve, reject) =>{
+        if (!fs.existsSync(directory)) {
+            mkdirp.sync(directory, (err) => {
+                if (err) {
+                    console.log(err);
+                    rejec(err);
+                }
+                else {
+                    console.log('Created directory at ', directory);
```
pshapiro4broad: I assume this log is for debugging, can it be removed now?

electron_app/helpers/_handleDisk.js
```diff
@@ -0,0 +1,27 @@
...
+    let data = {
+            'path': directory,
+            'item': item
+        };
+    return new Promise((resolve, reject) =>{
+        if (!fs.existsSync(directory)) {
+            mkdirp.sync(directory, (err) => {
+                if (err) {
+                    console.log(err);
+                    rejec(err);
```
pshapiro4broad: Does this compile? It looks like this should be `reject(err)` not `rejec(err)`.

tadeork: Yes, it's wrong but for some reason is compiling. I'll fix it.

electron_app/helpers/_handleDisk.js
tadeork: Sure.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
tadeork: This is because in FileExplorerComponent are the elements to be download. But in transferables-grid is the actual file list to download.

electron_app/helpers/_handleDisk.js
tadeork: About the same, what do you think of lines 18 and 23 in the DownloadManager?

electron_app/helpers/_handleDisk.js
```diff
@@ -0,0 +1,27 @@
+const fs = require('fs');
+const mkdirp = require('mkdirp');
+
+const handleFolder = (directory, item) => {
+    let data = {
```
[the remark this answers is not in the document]
tadeork: Yes, it's a best approach.

electron_app/DownloadManager.js
```diff
@@ -0,0 +1,55 @@
+const mtd = require('zeltice-mt-downloader');
+const path = require('path');
+const os = require('os');
+
+const handleFolder = require('./helpers/_handleDisk');
+
+let downElements = [];
+
+let downloadManager = (items, access_token) => {
+    let filePath = '';
```
pshapiro4broad: It looks like this code and `_handleDisk.js` uses four space indent, but other files use two space indent. I think we had decided on two space indent but can't recall for sure.

electron_app/DownloadManager.js
```diff
@@ -0,0 +1,55 @@
...
+    let promiseArr = [];
+    items.forEach(item => {
+        if (item.preserveStructure) {
+            const structurePath = item.path.substring(item.path.lastIndexOf('/'), 0);
+            promiseArr.push(handleFolder(path.join(item.destination, structurePath), item));
+        } else {
+            promiseArr.push(new Promise((res, rej) => { res({ 'path': item.destination, 'item': item }); }));
+        }
+    });
+    Promise.all(promiseArr).then(result => handleDownloads(result, options));
```
pshapiro4broad: Is the code using promises here to avoid having all files fail if one file fails to create a directory?

electron_app/helpers/handleDisk.js
```diff
@@ -0,0 +1,19 @@
+const fs = require('fs');
+const mkdirp = require('mkdirp');
+
+const handleFolder = (directory) => {
+    return new Promise((resolve, reject) =>{
+        if (!fs.existsSync(directory)) {
+
+            mkdirp.sync(directory, (err) => {
+                if (err) {
+                    console.log(err);
```
pshapiro4broad: You should test this, but I don't think `console.log()` will work in this context. It's for debugging only, since we'll need user-facing error reporting for this error, but if it doesn't work it will be confusing if it fails but doesn't produce a message.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
pshapiro4broad: OK, I understand. The models aren't quite correct yet; the list of items being downloaded is really a global collection that the transferables grid displays and observes for changes. When an item is queued for download (by closing the file modal dialog in `FileExplorerComponent.selectionDone()`) it should be added to the collection of items being downloaded. The work to start the download occurs outside this UI code, but this code can observe the collection so it knows to update its view of the data being downloaded.

It's similar to what's done here, but with the current code there's no way to know what elements in `filesDatabase` are already downloading and which ones haven't been started yet.

electron_app/DownloadManager.js
```diff
@@ -0,0 +1,50 @@
...
+            if (err) console.error(err);
+            else {
+                console.log('Download Complete');
+            }
+        }
+    };
+
+    let promiseArr = [];
+    items.forEach(item => {
+        if (item.preserveStructure) {
```
pshapiro4broad: With this change, I think you can remove `FilesDatabase.destinationChange` and `FilesDatabase.preserveStructureChange`

electron_app/DownloadManager.js
tadeork: Yes, I miss that in the .JS files.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
tadeork: Yes, I think I get you. I think this will be fixed in a specific story. When this changes also will changes some lines in the DownloadManager that take care of avoid the repetition of files to download.

PR #17, opened 2017-11-29
DSDEGP 1789- Add Total Size of Selected Data

wloforte:
### Description

_Add size pipe for tree table columns and add Total Size of Selected Files._

----

### Checklist

* Review cycle:


src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -146,6 +157,7 @@ export class FileExplorerComponent implements OnInit {
     });
     this.selectedFiles = newSelection;
     this.countFiles();
+    this.sumSize();
```
pshapiro4broad: Instead of duplicating the code that calls these functions, there should be a single function that updates both totals that's called from each method.

Or, better yet, the code should observe `selectedFiles` and update both counts if the selected files changes.

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -97,6 +105,7 @@ export class FileExplorerComponent implements OnInit {
   deleteNode(node: TreeNode) {
     node.parent.children = node.parent.children.filter(n => n.data !== node.data);
     this.countFiles();
+    this.sumSize();
```
pshapiro4broad: Looking at this change, it looks like `deleteNode()` is broken in that it could leave a node selected that's been deleted. Since we don't support delete this isn't the worst problem, but I think we should either fix deleteDelete or remove it until we have a use for it.

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -72,14 +69,25 @@ export class FileExplorerComponent implements OnInit {
     });
   }
 
+  sumSize() {
+    this.totalSize = 0;
+    this.selectedFiles.forEach(f => {
+      if (f.data.type === 'File') {
+        this.totalSize += parseFloat(f.data.size);
```
pshapiro4broad: Instead of calling `parseFloat()` every time the selection changes, why not store `size` as a `number` and call `parseFloat()` in the code that builds a `FileData`?

wloforte: Good catch, I fixed this issues. Let me know If there's something else I need to change. Thanks!

PR #23, opened 2017-12-13
Lf-DSDEGP-1747-Preflight Copy - Disk space and write permissions verifications

LeoForconesi:
### Description

- _The main goal, is to check disk space availability when a user attempts to download files, and display a friendly error message for this._

- _Secondarily, this code handles user's write permissions warning when he attempts to download a file and the destination folder, lacks of write permissions._

_**Motivation:** Due the files size to be downloaded from FC-explorer, it is necessary to handle disk's space possible errors, and give the user the possibility to recover from it._
_**Jira Ticket**, https://broadinstitute.atlassian.net/browse/DSDEGP-1747_

----

### Checklist

* Review cycle:


electron_app/DownloadManager.js
```diff
@@ -2,22 +2,29 @@ const mtd = require('zeltice-mt-downloader');
...
-        processDownload(access_token, item, result)
-      );
-    } else {
-      processDownload(access_token, item, item.destination);
-    }
+  return new Promise((res, rej) => {
+    let filePath = '';
+    const options = setHeader(access_token);
+    handleDiskSpace(items).then(saveConditions => {
+      if (saveConditions.diskSpace && saveConditions.allowed) {
```
pshapiro4broad: Since the body is only executed when both these cases are true, and since this is the only use of `handleDiskSpace()`, why not change it to not call the `then()` method at all if either condition is false? That would avoid the need to return boolean results from this method.

electron_app/helpers/handleDisk.js
```diff
@@ -15,4 +21,41 @@ const handleFolder = (directory) => {
...
+      totalSize = totalSize + item.size;
+    });
+    if (process.platform == 'win32') {
+      drive = drive.slice(0, 1);
+    }
+    diskspace.check(drive, function (err, diskSpace) {
+      canWrite(destinationPath, (err, isWritable) => {
+        saveConditions.enoughSpace = hasEnoughSpace(diskSpace.free, totalSize);
+        saveConditions.allowed = isWritable;
+        res(saveConditions);
```
pshapiro4broad: The change I describe above would be implemented like this:
```
  if (hasEnoughSpace(diskSpace.free, totalSize) && isWritable) {
    res(saveCondition);
  }
```

electron_app/helpers/handleDisk.js
```diff
@@ -15,4 +21,41 @@ const handleFolder = (directory) => {
...
+      });
+    })
+  });
+};
+
+const hasEnoughSpace = (localSpace, itemsSize) => {
+  if (localSpace > itemsSize) {
+    return true;
+  }
+  return false;
```
pshapiro4broad: this can be written as
```
  return localSpace > itemsSize;
```

electron_app/DownloadManager.js
LeoForconesi: I understand what you mean. Although there's a reason why the method returns an object with two booleans, which is that we need some type of flag to identify the status of each conditions (free disk space, and write permissions) so we can later display its proper warning in the UI, if you take a look to `main.js` here we send this boolean object to Angular's part. More specifically to `file-modal.ts`.

pshapiro4broad: OK. Since `rej` was unused I didn't see how `saveConditions` would be returned in the negative case. Another approach would be to return an error string since either error condition is a failure and it doesn't matter to the caller which failure it is.

pshapiro4broad:
I've built this and tried testing it and wasn't able to get an error message. I created directory, `/tmp/test` and changed it to read only using `chmod -w test` (and verified that `touch test/foo` produced a Permission denied error) and used that as my download destination. When I click the Start Download button, nothing happens. The dialog isn't dismissed and no error is shown. I checked the javascript console in electron and didn't see any errors there.

src/app/file-manager/file-modal/file-modal.component.ts
```diff
@@ -1,43 +1,104 @@
...
+    this.transferablesGridComponent.startDownload().then(
+      status => {
+        this.msgs = [];
+        if (!status.diskSpace) {
+          this.createWarningMsg('Sorry, there is not enough space on disk.');
+        } else if (!status.allowed) {
+          this.createWarningMsg('Sorry, you don\'t have the proper permissions to access that location.');
+        } else {
+          this.store.dispatch(new Downloadables.RemoveByStatus(Status.VERIFYING));
+          this.setItems(Status.DOWNLOADING);
```
pshapiro4broad: Instead of doing it this way, could you change the state of the items in the list from `VERIFYING` to `DOWNLOADING`? That seems like a better way to model the state of the transfer items.

src/app/file-manager/file-modal/file-modal.component.ts
```diff
@@ -1,43 +1,104 @@
...
 export class FileModalComponent  {
+  @Output('done') done: EventEmitter<any> = new EventEmitter();
 
   preserveStructure = true;
   directory = 'Choose Directory...';
   isValid = false;
+  dataFile: Item;
+  files: TreeNode[];
+  selectedFiles: TreeNode[] = [];
+  selectedFile: TreeNode;
```
pshapiro4broad: Both `selectedFiles` and `selectedFile` appear to be unused

src/app/file-manager/file-modal/file-modal.component.ts
```diff
@@ -1,43 +1,104 @@
...
   templateUrl: './file-modal.component.html',
   styleUrls: ['./file-modal.component.scss']
 })
 export class FileModalComponent  {
+  @Output('done') done: EventEmitter<any> = new EventEmitter();
 
   preserveStructure = true;
   directory = 'Choose Directory...';
   isValid = false;
+  dataFile: Item;
```
pshapiro4broad: It looks like `dataFile` is only used in `setItems()`, if so it should be changed to a local variable.

src/app/file-manager/services/register-download.service.ts
```diff
@@ -1,16 +1,22 @@
...
 @Injectable()
 export class RegisterDownloadService {
-
   constructor(private electronService: ElectronService) { }
 
-  public startDownload(items: Item[]) {
+  public startDownload(items: Item[]): Promise<boolean> {
     if (this.electronService.isElectronApp) {
-      this.electronService.ipcRenderer.send('start-download', items, SecurityService.getAccessToken());
+      this.electronService.ipcRenderer.removeAllListeners('disk-space');
```
pshapiro4broad: What is this doing? I don't see `'disk-space'` used elsewhere in the code.

src/app/file-manager/services/register-download.service.ts
```diff
@@ -1,16 +1,22 @@
...
+      return new Promise<boolean>((resolve, reject) => {
+        this.electronService.ipcRenderer.send('start-download', items, SecurityService.getAccessToken());
+        this.electronService.ipcRenderer.on('system-verification', (event, res) => {
```
pshapiro4broad: This should use the constants in `environment.js` instead, if possible.

electron_app/DownloadManager.js
```diff
@@ -2,22 +2,29 @@ const mtd = require('zeltice-mt-downloader');
...
-  items.forEach(item => {
-    if (item.preserveStructure) {
-      const structurePath = item.path.substring(item.path.lastIndexOf('/'), 0);
-      handleFolder(path.join(item.destination, structurePath)).then(result =>
-        processDownload(access_token, item, result)
-      );
-    } else {
-      processDownload(access_token, item, item.destination);
-    }
+  return new Promise((res, rej) => {
```
pshapiro4broad: Now that I've looked at this more, I'm not sure that this is the correct place to do this work. We want to check for the error case before the download begins, not after. The best scenario would be to report the error before the dialog is dismissed, e.g. when computing `FileModalComponent.isValid`. If not there, it should be done in `FileModalComponent.startDownload()` before it calls `transferablesGridComponent.startDownload()`. That way you can avoid adding and then removing items from the transfer list.

electron_app/helpers/handleDisk.js
```diff
@@ -1,18 +1,49 @@
...
+};
 
-      mkdirp.sync(directory, (err) => {
-        if (err) {
-          reject(err);
+const handleDiskSpace = (destination, totalFileSize) => {
+    var drive = destination;
+    Conditions = {error: Boolean, errorMessage: String};    
+    return new Promise((res, rej) => {
+        if (process.platform == 'win32') {
```
pshapiro4broad: This should use `===`, not `==`, correct? I thought we had a lint for this; but now see that this is a regular .js file not a .ts file. Can this code be written as .ts as well? I think that would be better, for readability and maintenance.

electron_app/helpers/handleDisk.js
```diff
@@ -1,18 +1,49 @@
...
+        }
+        resolve(directory);
+    });
+};
 
-      mkdirp.sync(directory, (err) => {
-        if (err) {
-          reject(err);
+const handleDiskSpace = (destination, totalFileSize) => {
+    var drive = destination;
```
pshapiro4broad: I think `let` is preferred over `var` here

electron_app/DownloadManager.js
LeoForconesi: i have changed the return object to a more generic one, instead of  `[hasEnoughSpace :Boolean, allowed: Boolean}` this is now `{hasError: Boolean, errorMessage: String}`. That way, there would be only one flag for an error and a better handling of the specific error. 
Regarding `rej()` function for the promise, i'm adding a try/catch block to get any error coming from there, and login it as a console.error.

src/app/file-manager/file-modal/file-modal.component.ts
LeoForconesi: This has been changed, so now there is no necessity to handle this status from the store.

src/app/file-manager/file-modal/file-modal.component.ts
LeoForconesi: These variables have been removed.

src/app/file-manager/file-modal/file-modal.component.ts
LeoForconesi: `dataFile` is now a local variable.

src/app/file-manager/services/register-download.service.ts
LeoForconesi: This is now fixed.

main.js
```diff
@@ -36,7 +37,7 @@ app.on('ready', function () {
   }
   // Show dev tools
   // Remove this line before distributing
-  // win.webContents.openDevTools()
+  win.webContents.openDevTools()
```
pshapiro4broad: Is there a way to wrap this line in something like `if (running production) then`?

electron_app/DownloadManager.js
LeoForconesi: All disk verification were changed to be performed separately from any download action. 
First, check disk conditions, if everything ok, download else, show error message.

electron_app/helpers/handleDisk.js
LeoForconesi: I've corrected `===` operator. 
On the other hand, what you mentioned regarding `.js` to `.ts` files sounds reasonable, so yes, please lets create a Jira ticket to address this improvement.

electron_app/helpers/handleDisk.js
LeoForconesi: Fixed!

main.js
```diff
@@ -79,6 +80,17 @@ app.on('ready', function () {
   ipcMain.on(constants.IPC_START_DOWNLOAD, (event, items, access_token) => {
     downloadManager(items, access_token);
   });
+
+  ipcMain.on(constants.IPC_VERIFY_BEFORE_DOWNLOAD, (event, destination, totalFilesSize) => {
```
pshapiro4broad: I'm curious why you're using IPC to implement this API instead of calling `handleDiskSpace()` directly.

src/app/file-manager/services/register-download.service.ts
```diff
@@ -1,16 +1,33 @@
...
   public startDownload(items: Item[]) {
     if (this.electronService.isElectronApp) {
       this.electronService.ipcRenderer.send('start-download', items, SecurityService.getAccessToken());
     }
   }
+
+  public verifyDisk(destination: String, totalfilesSize: Number): Promise<DiskStatus> {
+    return new Promise(result => {
+
+      this.electronService.ipcRenderer.removeAllListeners('return-disk-response');
```
pshapiro4broad: Can you use the constants in `environment.js` here?

pshapiro4broad: I don't think this is the correct way to have an IPC that sends a response back; you shouldn't need another IPC message for a response. It looks like using a Promise would work here: https://www.npmjs.com/package/electron-ipc

LeoForconesi: This library hasn't been updated for more than a year, and we cannot access its code from github, so we think it isn't convenient to implement this.

main.js
LeoForconesi: Since Electron uses Chromium for displaying web pages, Chromium's multi-process architecture is also used. Each web page in Electron runs in its own process, which is called the `renderer process`. 
So, in normal browsers, web pages usually run in a sandboxed environment and are not allowed to access to native resources. Electron users, however, have the power to use `Node.js` APIs in web pages allowing lower level operating system interactions.
In this case `handeDiskSpace()` uses a library to check free disk space that runs commands for specific operative system, being necessary to run this method on a lower level, which means that we need to go through IPC implementation. 

Here's some reference info:
https://electronjs.org/docs/tutorial/quick-start  -> the first part of this article explains diferences between renderer processes and main process.

https://www.npmjs.com/package/diskspace -> this is node library that we are using to check disk space, you can find a link to github repo if you feel like to check the code, its not very extensive, but here you can see why we need to use lower operative system level.

pshapiro4broad: Thanks for the explanation, that makes sense.

src/app/file-manager/services/register-download.service.ts
pshapiro4broad: Oh, I thought this was part of electron, not an add-on. Can you use `ipcRender.once()` to avoid having to remove the listener?

LeoForconesi: Oh yeah! `ipcRender.once()` works smoothly. Nice! I'll update this change.

PR #37, opened 2018-01-02
DSDEGP-1912 Alpha release refinements

nlopezzalba:
### Description

-UI says "Show detailed download list" but it should be "Show detailed transfer list".
In the "Status" page, the operation "Show detailed download list" operates as a button but looks like a regular block of text. Instead it should have button appearance. Since it's a toggle button that acts as a disclosure control, using a progressive disclosure element such as a triangle would be appropriate. Another possibility is to always show the detail view instead of having an option to hide it.
*Our solution approach was adding an arrow at the left of the text and a hover effect as well as changing the font size*

-UI uses "(s)" in many places to indicate possible plurals. In cases where the number isn't known (such as "Select files(s) to upload"), the plural form should be used. In cases where the size is known (such as "Selected: 0 file(s)"),the correct form should be used based on the known number.

-In upload view, the size of directories is incorrect. We may not be able to compute the total directory size lazily (e.g. without completely reading all subdirectory contents). Having an incorrect size seems worse then having no size at all; if this can't be fixed to be accurate (and efficient), the UI should show '-' or some other indicator to show that no size is known. The Total Size in the Selected view at the bottom should also be updated so it doesn't show an incorrect value, in the case where a user selects a directory.
*In upload view folder's size won't be shown, since we are using lazy loading we don't actually know folder size until it's expanded*

In Progress Status page, "Cancel" buttons for download and upload will be disabled until a transfer begins. So it can be clicked by the user only when a transfer is in progress.

----

### Checklist

- [X] **Submitter**: JIRA ticket checks:
* Review cycle:
- [X] **Submitter**: Verify all tests go green, including CI tests


src/app/file-manager/transferables-grid/transferables-grid.component.ts
```diff
@@ -90,11 +92,21 @@ export class TransferablesGridComponent implements OnInit, AfterViewInit {
     this.statusService.updateDownloadProgress().subscribe(data => {
       this.zone.run(() => {
         this.generalProgress = data;
+        if (data === 100) {
+          this.downloadInProgress = false;
+        } else {
+          this.downloadInProgress = true;
+        }
```
pshapiro4broad: would be simpler to write this as 
```
this.downloadInProgress = data !== 100;
```
Also, we need to get rid of magic numbers in this code, such as `100`, but that can wait for a later cleanup.

nlopezzalba: Thanks, we will refactor this in a later cleanup.

PR #44, opened 2018-01-18
DSDEGP-1965-Handle case where a downloaded file conflicts with an existing file

LeoForconesi:
### Description

_This code will check a file destination folder to see if an existing file with the same name exists there before download any file._
_If the file exists, it will add a number between parenthesis to the new copy_
_**Motivation:** in the previous version of the app the file was replaced with the new file._
_**Jira Ticket:** [DSDEGP-1965](https://broadinstitute.atlassian.net/browse/DSDEGP-1965)_

----

### Checklist

* Review cycle:
- [x] **Submitter**: Delete branch after merge


electron_app/DownloadManager.js
```diff
@@ -1,25 +1,48 @@
...
+        item.name = fileName;
+        processItem(item, access_token, electronWin);
+      }
+    } while (fileExists);
   });
 };
 
+const fileExistsInDestination = (item, nameCheck) => {
+  fileExists = item.preserveStructure ?
+    fileAlreadyExists(path.join(item.destination + '/' + item.path.substring(item.path.lastIndexOf('/'), 0) + '/' + nameCheck))
```
pshapiro4broad: I'm pretty sure this is not the right way to use `path.join()`; you want to avoid hardcoding `/` and pass in each path element as a separate argument, e.g. `path.join(item.destination, item.path.substring(...), nameCheck)`

electron_app/DownloadManager.js
```diff
@@ -1,25 +1,48 @@
...
+    : fileExists = fileAlreadyExists(path.join(item.destination + '/' + nameCheck));
+};
+
+const processItem = (item, token, win) => {
+  if (item.preserveStructure) {
+    const structurePath = item.path.substring(item.path.lastIndexOf('/'), 0);
```
pshapiro4broad: It would be nice if there was a way to avoid testing for `preserveStructure` twice. It also duplicates the way the path is constructed. Since the relative path doesn't change on each loop, only the filename, the destination directory can be computed once before the `do` loop on line 15 above. That would avoid the need to check `preserveStructure` in the loop.

electron_app/DownloadManager.js
LeoForconesi: Fixed!

electron_app/DownloadManager.js
LeoForconesi: Thanks for the advice, I removed that verification from the loop, and item's destination path is only constructed once, changing each file's name if necessary.

electron_app/DownloadManager.js
```diff
@@ -1,29 +1,48 @@
...
+    const destination = item.preserveStructure ?
+      path.join(item.destination, item.path.substring(item.path.lastIndexOf('/'), 0)) : item.destination;
+    do {
+      fileExists = fileAlreadyExists(path.join(destination, fileName));
+      if (fileExists) {
+        count++;
+        fileName = item.name.substring(0, item.name.indexOf('.')) + '(' + (count) + ')' +
+          item.name.substring(item.name.indexOf('.'));
+      } else {
+        item.name = fileName;
```
pshapiro4broad: One last thing here, I think it would be clearer to move this `else` clause outside the `do/while` loop since it only executes once, when the loop is complete. Having it inside the `do/while` makes it seem like it might be executed more than once.

electron_app/DownloadManager.js
```diff
@@ -1,29 +1,48 @@
...
+        item.destination = destination;
+        processItem(item, access_token, electronWin);
+      }
+    } while (fileExists);
   });
 };
 
+const processItem = (item, token, win) => {
+  if (item.preserveStructure) {
+    handleFolder(item.destination).then(result => {
```
pshapiro4broad: You could avoid overwriting `item.destination` above by passing in the destination directory as an argument to `processItem()`. I think you can also remove the check for `preserveStructure` here. If `preserveStructure` is false than `handleFolder()` won't need to create a directory at all, but it doesn't hurt to call it anyway.

pshapiro4broad (approved):
👍 Just one last round of comments

electron_app/DownloadManager.js
LeoForconesi: Fixed!

pshapiro4broad (approved):
Looks good.

electron_app/DownloadManager.js
LeoForconesi: I had to change handleFolder return. Before it was a promise, now its a callback, this is due its synchronous behavior caused some issues on multiple downloads with the exact same name and extension.

PR #45, opened 2018-01-22
DSDEGP-1921 - Download screen takes a very long time to load when user has many workspaces/files

vvicario:
### Description

Download screen takes a very long time to load when user has many workspaces/files

----

### Checklist

- [ ] **Submitter**: Include the JIRA issue number in the PR description
* Review cycle:


src/app/file-manager/file-explorer-upload/file-explorer-upload.component.ts
```diff
@@ -231,7 +229,10 @@ export class FileExplorerUploadComponent implements OnInit {
               progress: 0,
               status: ItemStatus.PENDING,
               transferred: 0,
-              remaining: 0
+              remaining: 0,
+              bucketName: '',
+              delimiter: '',
+              prefix: ''
```
pshapiro4broad: Instead of duplicating the default settings for these fields, can you add a constructor to `Item` that handles this?

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -74,25 +74,28 @@ export class FileDownloadModalComponent  {
...
-          size: file.data.size,
-          created: file.data.updated,
-          updated: file.data.updated,
-          icon: file.data.type === 'Folder' ? 'folder' : 'cloud',
+          id: file.id,
+          name: file.name,
+          size: file.size,
+          created: file.updated,
+          updated: file.updated,
+          icon: file.type === 'Folder' ? 'folder' : 'cloud',
```
pshapiro4broad: It looks like `Folder` will never occur because of the call to `.filter()` above.

pshapiro4broad:
Now that we have a paginated view, does Select All only select items that are currently in view? If so, maybe we don't need Select All anymore since clicking the checkbox at the top would have the same behavior. Or, we could get rid of the checkbox. We already have an issue where the checkbox is selected even if no items in view are selected, since it reflects the state of the entire selection, not the current page, which seems a bit confusing to me.

@nlopezzalba what do you think?

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -74,104 +64,146 @@ export class FileExplorerComponent implements OnInit {
...
-    node.expanded = isExpand;
-    if (node.children) {
-      node.children.forEach(childNode => {
-        this.expandRecursive(childNode, isExpand);
+  /** Selects all rows if they are not all selected; otherwise clear selection. */
+  masterToggle() {
+    this.isAllSelected() ?
+      this.clearSelection() :
+      this.bucketObjects.data.forEach(row => {
+       if (row.type === 'File') {
```
pshapiro4broad: wrong number of spaces in indent

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -74,104 +64,146 @@ export class FileExplorerComponent implements OnInit {
...
-
     });
   }
 
   disableButton() {
-    return this.downloadInProgress || this.fileCount <= 0;
+    return this.downloadInProgress || this.selection.selected.length === 0;
+  }
+
+  isSelected(event, row) {
```
pshapiro4broad: This seems like a strange name for this method. How about `toggleRow()`?

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -74,104 +64,146 @@ export class FileExplorerComponent implements OnInit {
...
+      this.selection.deselect(selectedItem);
+    }
+    this.countFiles();
+  }
+
+  getWorkspaceName(bucketName) {
+    let workspaceName = '';
+    FirecloudApiService.workspaces.forEach((value: string, key: string) => {
+      if (value === bucketName) {
+        workspaceName = key;
```
pshapiro4broad: Once you find the matching workspace, you can `break;` to avoid checking more workspaces.

nlopezzalba:
@pshapiro4broad I thought we got rid of "Select all" funcionality, let's provide the user the table header checkbox only (the idea is to make it work properly during next sprint).

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -199,7 +205,9 @@ export class FileExplorerComponent implements OnInit {
 
   getDelimiter(selectedFolder) {
     const path = this.breadcrumbPath.substring(this.breadcrumbPath.indexOf(this.DELIMITER), this.breadcrumbPath.length);
-    return path.substring(1, path.lastIndexOf(selectedFolder)) + selectedFolder;
+    const indexSelectedFolder = path.lastIndexOf('/' + selectedFolder + '/') !== -1 ? path.lastIndexOf('/' + selectedFolder + '/') + 1
```
pshapiro4broad: This would be clearer if `path.lastIndexOf('/' + selectedFolder + '/')` was stored in a variable, so it isn't duplicated in the code.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -76,27 +76,9 @@ export class FileDownloadModalComponent  {
...
-          status: ItemStatus.PENDING,
-          transferred: 0,
-          remaining: 0,
-          bucketName: '',
-          delimiter: '',
-          prefix: ''
-        };
+        const dataFile: Item = new Item(file.id, file.name, file.updated, file.created,
+          file.size, file.mediaLink, file.path, this.directory,
+          Type.DOWNLOAD, ItemStatus.PENDING, '', '', '', this.preserveStructure);
```
pshapiro4broad: I think you can improve this slightly more by making a constructor specific to file upload/download that omits the arguments `Type.DOWNLOAD, ItemStatus.PENDING, '', '', ''` since they are the same in both cases. Not sure if you could replace the `file` arguments as well (e.g. pass `file` for download and pass `file.data` for upload, but that would help too.

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -18,23 +17,18 @@ import { FilesDataSource } from '@app/file-manager/dbstate/files-datasource';
   styleUrls: ['./file-explorer.component.scss']
 })
 export class FileExplorerComponent implements OnInit {
+  IS_HOME: Boolean = false;
```
pshapiro4broad: If this isn't a constant it should probably be named like a regular field, e.g. `isHome`

pshapiro4broad (approved):
One last round of comments, 👍 once they are addressed

src/app/file-manager/file-download-modal/file-download-modal.component.ts
vvicario: We can't create multiple constructors for each type because multiple implementations are not allowed. It would be nice to send only one parameter with Item argument but we are using the same constructor to create   files, and folders in Bucket Service, in this case we don't have the specific item so we have to create it before call the constructor.

pshapiro4broad: OK. Writing a static factory method is another possible solution.

PR #46, opened 2018-01-23
DSDEGP-1935 Test Coverage

tadeork:
### Description

_Increment the test coverage, in this case LoginComponent, LoginService and FireCloudService._
_Also added minor change made to main.js to not show the Electron default message **'Loading...'**_
_JIRA story [DSDEGP-1935](https://broadinstitute.atlassian.net/browse/DSDEGP-1935)._

----

### Checklist

* Review cycle:
- [x] **Submitter**: Delete branch after merge


pshapiro4broad (approved):
I have a couple comments, 👍 once responded to.
 
This looks like a good start. I was hoping to see tests of the download or upload pages, but having the login covered well seems like a good place to start from.

src/app/file-manager/file-manager.module.ts
```diff
@@ -62,6 +68,17 @@ import { MatDialog } from '@angular/material';
     StatusService,
     FilesDatabase,
     FilesService,
+    {
+      provide: LoginService,
+      deps: [ElectronIpcService, Store, FirecloudService],
+      useFactory(electronIpc: ElectronIpcService, store: Store<any>, firecloudService: FirecloudService) {
+        if (environment.testing) {
```
pshapiro4broad: Is there no way to avoid doing this every place the service is used? I would expect that there would be a way to do this in the service definition code, not at the point were the service is imported. This way seems like you'd end up repeating this test in different modules.

src/app/file-manager/login/login.component.spec.ts
```diff
@@ -1,45 +1,115 @@
...
+  describe('Log the user', () => {
+    it('should log', () => {
+      component.logIn();
+      expect(component.hasError).toBeFalsy();
+    });
+  });
+
+  describe('Show errors', () => {
+    it('should show messages for 401 error code', () => {
+      component.showError(401);
```
pshapiro4broad, line 124: This is OK but it would be better to drive the error from the service layer, e.g. have the login request return the http status code and then verify that the error show in the view is correct.

tadeork: Sure, I'm looking how to implement that.

tadeork:
@pshapiro4broad I think we can create a story to address the tests of the download or upload pages

src/app/file-manager/file-manager.module.ts
tadeork: Yes, that's correct. In fact we don't need to do this any more at least for the tests.

src/app/file-manager/login/login.component.spec.ts
tadeork: This is taking me too much time to solve. I'll tackle this in the next approach with the following tests.

PR #47, opened 2018-01-29
DSDEGP-2001 Select all only after response is complete

vvicario:
DSDEGP-2001 Select all only after response is complete

_Please explain the changes you made here._
_Explain the **motivation** for making this change. What existing problem does the pull request solve?_
_Mention any JIRA issues fixed, addressed or otherwise related to this pull request, including issue numbers or hard links for issues in other repos._

----

### Checklist

- [ X] **Submitter**: Include the JIRA issue number in the PR description
- [ X] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
* Review cycle:


src/app/file-manager/services/bucket.service.ts
```diff
@@ -9,6 +9,7 @@ import { Type } from '@app/file-manager/models/type';
 export class BucketService {
 
   static delimiter: string;
+  static isResponseComplete = false;
```
pshapiro4broad: Why is this `static`? I know we only have a single 'download' view but it doesn't seem correct to make this state static when it could be a field on `BucketService`, and the file explorer component already has a field `bucketService`.

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -161,6 +163,7 @@ export class FileExplorerComponent implements OnInit {
   }
 
   getBucketObjects(bucket: String, delimiter: String, workspace: String) {
+    BucketService.isResponseComplete = false;
```
pshapiro4broad: Can this be done in  `BucketService.getBucketData()`?

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -119,13 +119,15 @@ export class FileExplorerComponent implements OnInit {
   }
 
   public goRoot() {
+    BucketService.isResponseComplete = false;
```
pshapiro4broad: This is a bit ugly, as the code is reusing the bucket service property to track a different state, the state of the `getUserFirecloudWorkspaces()` call. It would be better for this class to have its own field and then define `isResponseComplete()` below as the combination of both states.

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -209,6 +212,10 @@ export class FileExplorerComponent implements OnInit {
     return path.substring(1, indexSelectedFolder) + selectedFolder;
   }
 
+  isResponseComplete() {
```
pshapiro4broad: It might be clearer to call this method something like `dataFinishedLoading()`, that would make the intent of the HTML clearer.

vvicario:
@pshapiro4broad I worked on your feedback. Please let me know if you have any other comment.

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -161,6 +162,7 @@ export class FileExplorerComponent implements OnInit {
   }
 
   getBucketObjects(bucket: String, delimiter: String, workspace: String) {
+    this.bucketService.setIsResponseComplete(false);
```
pshapiro4broad: Can this be done in `BucketService.getBucketData()`?

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -120,6 +120,7 @@ export class FileExplorerComponent implements OnInit {
 
   public goRoot() {
     if (!this.isHome) {
+      this.firecloudService.setIsResponseComplete(false);
```
pshapiro4broad: Can this be done in `FirecloudService.getUserFirecloudWorkspaces()`?

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -209,6 +211,16 @@ export class FileExplorerComponent implements OnInit {
     return path.substring(1, indexSelectedFolder) + selectedFolder;
   }
 
+  dataFinishedLoading() {
+    let dataFinishedLoading = false;
+    if (this.isHome && this.firecloudService.getIsResponseComplete()) {
```
pshapiro4broad: Is it necessary to check `this.isHome` here? I would expect this function to look like:
```
dataFinishedLoading() {
  return !(this.firecloudService.getIsResponseComplete() || this.bucketService.getIsResponseComplete());
}
```

pshapiro4broad (approved):
One more round; I think you can get rid of the `setIsResponseComplete()` methods if you move the logic to `set(false)` into the service methods.

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -213,13 +211,19 @@ export class FileExplorerComponent implements OnInit {
...
 
   dataFinishedLoading() {
-    let dataFinishedLoading = false;
-    if (this.isHome && this.firecloudService.getIsResponseComplete()) {
-      dataFinishedLoading = true;
-    } else if (this.bucketService.getIsResponseComplete()) {
-      dataFinishedLoading = true;
-    }
-    return dataFinishedLoading;
+    return this.firecloudService.getIsResponseComplete() || this.bucketService.getIsResponseComplete();
```
pshapiro4broad: I think this is `&&` not `||` since both must be complete.

vvicario: Both must not be completed at the same time, if we are in the first page we only call firecloud to get workspaces data, if we move to another folder we call bucket service to get bucket info but this occur in different moments. I just added isHome condition to check if we need to verify one or another response

pshapiro4broad: OK, as long as we know which one the UI is currently waiting for.

pshapiro4broad (approved):

PR #51, opened 2018-02-01
DSDEGP-1906-Reimplement .js files as .ts; avoid using .js for new files

LeoForconesi:
### Description

_In this code you will find the re-implementation of the following files to Ts extension and format:_
- main.js
- uploadManager.js
- downloadManager.js

_**Motivation** The project now has both javascript and typescript files. It would be better to only use typescript, as it is more descriptive and has good lint support. In general we should be able to replace our non-configuration .js files as .ts files. In the future we should use .ts files over .js when possible._

**_Jira Story:_** [_DSDEGP-1906._](https://broadinstitute.atlassian.net/browse/DSDEGP-1906)

----

### Checklist

* Review cycle:


electron_app/DownloadManager.ts
```diff
@@ -0,0 +1,73 @@
+const Downloader = require('./Downloader');
+const path = require('path');
+const { handleFolder, fileAlreadyExists } = require('./helpers/handleDisk');
+const downloadStats = require('./helpers/downloadInfo').downloadStats;
+const handleEvents = require('./helpers/handleEvents');
+
+let filePath = '';
```
pshapiro4broad, line 7: The files are now .ts but it loks like they're not really converted completely yet. E.g. `filePath` and `fileExists` should be local variables and `allDownloads` should be a field in `DownloadManagerClass`. Also the `any`s should be migrated to use types.

pshapiro4broad:
While I appreciate the amount of effort involved in this change, I'm not convinced that this is adding enough benefit to offset the issue where the build output is written to the source tree.

I'll leave this up to you (@LeoForconesi) and the rest of the Belatrix team to decide, if you think it's a net benefit, then go ahead and merge it in. If not we can save the work as a .patch and add it to the JIRA to look at again later.

PR #57, opened 2018-03-05
Dg dsdegp 1997

dags:
### Description

We have implemented the selection of files and folders to download. 
Related stories : DSDEGP-1997 DSDEGP-1999
Know issues : 
1) when changing selection from header checkbox and the table includes many files or folders, may take some time and spinner is not working at that level. We are working to fix this.
2) Preflight may fail and render application is an unusable state in some conditions where selection includes a lot of folders. This is due to a huge amoutn of requests being launched at same time. We need to limit the concurrency of http requests. 

----

### Checklist

* Review cycle:


src/app/app.module.ts
```diff
@@ -12,6 +12,8 @@ import { AppRoutingModule } from './app-routing.module';
 import { AppComponent } from './app.component';
 import { NgxElectronModule } from 'ngx-electron';
 import { GoogleLoginService } from '@app/file-manager/services/login-google.service';
+// import { NgxSpinnerModule } from 'ngx-spinner';
+// import { NgxSpinnerService } from 'ngx-spinner';
```
pshapiro4broad: If the commented out lines in this file aren't needed can they be removed for now?

src/app/file-manager/breadcrumb/breadcrumb.component.ts
```diff
@@ -23,30 +28,43 @@ export class BreadcrumbComponent implements OnInit {
...
   homeClicked(evt) {
     this.home.emit('home');
   }
 
-  private processPath(fullpath: string) {
-    if (fullpath === null || fullpath === undefined) { return; }
-    fullpath = fullpath.substring(fullpath.length - 1, fullpath.length) === this.DELIMITER
-               ? fullpath.substr(0, fullpath.lastIndexOf(this.DELIMITER)) : fullpath;
-    const segs = fullpath.split(this.DELIMITER);
+  private processItems(steps: Item[]) {
```
pshapiro4broad: Not sure why this is `steps` and not `items`, I would expect it to be `items`.

src/app/file-manager/breadcrumb/breadcrumb.component.ts
```diff
@@ -23,30 +28,43 @@ export class BreadcrumbComponent implements OnInit {
...
-        label: s,
-        path: p
-      };
-      if (e.path !== '') {
+
+    if (steps === null || steps === undefined || steps.length <= 0) {
+      return;
+    }
+
+    steps.forEach(
```
pshapiro4broad: It looks like the type used here for `steps` doesn't match the parameter declaration `Item[]`. It also looks like this code refers to each element of the breadcrumb as a `step`. I haven't seen that term used before in this context, is this the usual name for this item? I think I've seen code call this element a `crumb` before.

src/app/file-manager/breadcrumb/breadcrumb.component.ts
```diff
@@ -23,30 +28,43 @@ export class BreadcrumbComponent implements OnInit {
...
+      (item, index) => {
+        const paths = item.prefix.split('/');
```
pshapiro4broad: The code defines `DELIMITER` above but it's not used at all in this class. Is this a case where it should be used?

src/app/file-manager/breadcrumb/breadcrumb.component.ts
```diff
@@ -23,30 +28,43 @@ export class BreadcrumbComponent implements OnInit {
...
+        // console.log(item, paths);
+        let label = item.path === (item.workspaceName + '/') ? item.name : paths.slice(-2, -1)[0];
```
pshapiro4broad: This needs a comment to explain what it's doing, why using -2, -1 is correct.

src/app/file-manager/breadcrumb/breadcrumb.component.ts
```diff
@@ -23,30 +28,43 @@ export class BreadcrumbComponent implements OnInit {
...
+        if (item.namespace !== '') {
+          label = item.namespace + ' | ' + label;
+        }
+
+        const path = item.path === (item.workspaceName + '/') ? item.name : item.path.split('/').slice(0, -1).join('/');
```
pshapiro4broad: This is a bit clearer than line 55 above but a comment would be nice here too.

src/app/file-manager/file-explorer/file-explorer.component.ts
```diff
@@ -1,37 +1,48 @@
...
-  displayedColumns = ['select', 'name', 'size', 'lastdate'];
-  dataSource: FilesDataSource | null;
-  selection = new SelectionModel<Item>(true, []);
+  displayedColumns = ['select', 'displayName', 'size', 'updated'];
+
+  dataSource = new MatTableDataSource([]);
+  @ViewChild(MatSort) sort: MatSort;
   @ViewChild(MatPaginator) paginator: MatPaginator;
+
+  breadcrumbItems: Item[];
```
pshapiro4broad, line 39: I'm confused why we have `breadcrumbItems` here as well as in the breadcrumb component above.

src/app/file-manager/breadcrumb/breadcrumb.component.html
```diff
@@ -2,7 +2,12 @@
...
   </li>
-  <li *ngFor="let segment of segments">
-    <a (click)="stepClicked($event, segment)">{{segment.label}}</a>
+  <li *ngFor="let segment of segments; let ix = index">
+    <a (click)="stepClicked($event, segment, ix)">{{segment.label}}</a>
   </li>
-</ul>
\ No newline at end of file
+</ul>
+<!-- <ul>
```
als364: Can these commented out lines be removed until they are needed?

src/app/file-manager/file-download-modal/file-download-modal.component.html
```diff
@@ -2,19 +2,33 @@ <h2 mat-dialog-title>Where do you want to download to?</h2>
...
 </mat-dialog-content>
 
-<mat-checkbox [checked]="preserveStructure" aria-label="Preserve folder structure" (change)="selectionChanged($event)" class="dialog-checkbox" disableRipple>Preserve folder structure</mat-checkbox>
+<mat-checkbox [checked]="preserveStructure" aria-label="Preserve folder structure" (change)="selectionChanged($event)" class="dialog-checkbox"
+  disableRipple>Preserve folder structure</mat-checkbox>
 <p-messages class="error-messages" [(value)]="msgs"></p-messages>
 <mat-dialog-actions>
-  <button mat-raised-button (click)="cancel()" class="button-secondary">Cancel</button>
-  <button mat-raised-button [disabled]="!isValid" (click)="startDownload()" class="button-primary">Start Download!</button>
+  <button mat-raised-button [disabled]="loading" (click)="cancel()" class="button-secondary">Cancel</button>
```
pshapiro4broad: Is this requirement a result of the implementation? I think that we want to allow users to cancel the preflight in the case where it takes a long time, so they can select fewer items. If the implementation prevents it for now, can you create a JIRA to add this feature later on?

src/app/file-manager/breadcrumb/breadcrumb.component.ts
```diff
@@ -23,30 +28,43 @@ export class BreadcrumbComponent implements OnInit {
...
-      if (e.path !== '') {
+
+    if (steps === null || steps === undefined || steps.length <= 0) {
+      return;
+    }
+
+    steps.forEach(
+      (item, index) => {
+        const paths = item.prefix.split('/');
+        // console.log(item, paths);
```
als364: Please remove commented out code.

src/app/file-manager/breadcrumb/breadcrumb.component.ts
```diff
@@ -23,30 +28,43 @@ export class BreadcrumbComponent implements OnInit {
...
+        const e = {
+          label: label,
+          path: path,
+          item: item
+        };
         this.segments.push(e);
-      }
-    });
+      });
+      // console.log(this.segments);
```
als364: Please remove commented out code.

src/app/file-manager/breadcrumb/breadcrumb.component.ts
```diff
@@ -23,30 +28,43 @@ export class BreadcrumbComponent implements OnInit {
...
+      (item, index) => {
+        const paths = item.prefix.split('/');
+        // console.log(item, paths);
+        let label = item.path === (item.workspaceName + '/') ? item.name : paths.slice(-2, -1)[0];
+        if (item.namespace !== '') {
+          label = item.namespace + ' | ' + label;
+        }
+
+        const path = item.path === (item.workspaceName + '/') ? item.name : item.path.split('/').slice(0, -1).join('/');
+        const e = {
```
als364: This variable name should be more descriptive.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -11,38 +10,172 @@ import { DownloadValidatorService } from '@app/file-manager/services/download-va
...
-        this.isValid = true;
+    this.filesMap = new Map();
+
+    if (localStorage.getItem('directory') !== null) {
+      this.directory = localStorage.getItem('directory');
+      this.isValid = true;
+    }
+    BucketService.delimiter = null;
+
+    // this.fileCountObservable.subscribe(
```
pshapiro4broad: If you want to leave in debugging statements, can you put them in a block like `if (environment === TEST)` instead of commenting them out.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -11,38 +10,172 @@ import { DownloadValidatorService } from '@app/file-manager/services/download-va
...
+
+  ngOnInit() {
+    this.processFilesToDownload();
+  }
+
+  processFilesToDownload() {
+
+    const patchedItems = this.data.selectedFiles
+      .filter(item => {
+        if (item.id === 'workspaces') {
```
pshapiro4broad: can simplify to `item => item.id !== 'workspaces'`

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -11,38 +10,172 @@ import { DownloadValidatorService } from '@app/file-manager/services/download-va
...
+
+  }
+
+  ngOnInit() {
+    this.processFilesToDownload();
+  }
+
+  processFilesToDownload() {
+
+    const patchedItems = this.data.selectedFiles
```
pshapiro4broad: Documentation would be helpful here. What is `patchedItems`? Why are the items sorted? etc.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -92,4 +229,77 @@ export class FileDownloadModalComponent  {
       summary: warnMessage,
     });
   }
+
+  // This function filters repeated folders when a parent
+  // is all-selected and there are sub-folders in the list
+  filterFolderItems(sortedItems): Item[] {
```
pshapiro4broad: typescript supports java-doc style comments which should be used to document methods, e.g.
```
    /**
     * This function filters repeated folders when a parent
     * is all-selected and there are sub-folders in the list
     */
```

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -11,38 +10,172 @@ import { DownloadValidatorService } from '@app/file-manager/services/download-va
...
+          if (item.path === '') {
+            return this.cloneItem(item);
+          }
+          return item;
+        }
+      );
+
+    const sortedItems = patchedItems.sort(
+      (itemA, itemB) => {
+        const pathA = itemA.path.toUpperCase();
```
pshapiro4broad: Can you use `localeCompare()` here? It looks like it will do case insensitive sort using `{'sensitivity' : 'base'}`

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -11,38 +10,172 @@ import { DownloadValidatorService } from '@app/file-manager/services/download-va
...
+    this.loadingFiles = true;
+    this.loadingFolders = 0;
+
+    cleanedItems
+      .forEach(
+        item => {
+          if (item.type === 'File') {
+            this.addToSelectedFiles(item, 'directly');
+          } else {
+            // while (this.loadingFolders > 5) {
```
pshapiro4broad: Not sure what this is doing, is it debugging code that can be deleted?

src/app/file-manager/file-download-modal/file-download-modal.component.ts
als364: Please remove commented out code.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -11,38 +10,172 @@ import { DownloadValidatorService } from '@app/file-manager/services/download-va
...
     public dialogRef: MatDialogRef<FileDownloadModalComponent>,
     public router: Router,
+    public bucketService: BucketService,
     @Inject(MAT_DIALOG_DATA) public data: any) {
-      if (localStorage.getItem('directory') !== null) {
-        this.directory = localStorage.getItem('directory');
-        this.isValid = true;
+    this.filesMap = new Map();
+
+    if (localStorage.getItem('directory') !== null) {
```
als364: Rather than retrieving from `localStorage` multiple times, this could be phrased as:
```
var directory = localStorage.getItem('directory')
if (directory !== null) {
    this.directory = directory
    this.isValid = true;
}
```

src/app/file-manager/file-explorer/file-explorer.component.ts
dags: breadcrumbItems array here holds the items:Item[] to be passed to bredcrumb components.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
als364: Please remove commented out code.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -11,38 +10,172 @@ import { DownloadValidatorService } from '@app/file-manager/services/download-va
...
+  }
+
+  isLoading() {
+    return this.loadingFiles || this.loadingFolders > 0;
+  }
+
+  addToSelectedFiles(item: Item, mode: string) {
+    const found: Item = this.selectedFiles.find(x => x.mediaLink === item.mediaLink);
+    if (found === undefined) {
+      // this.zone.run(() => {
```
als364: Please remove commented out code.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -11,38 +10,172 @@ import { DownloadValidatorService } from '@app/file-manager/services/download-va
...
+          this.loadingFolders--;
+        }
+      );
+  }
+
+  isLoading() {
+    return this.loadingFiles || this.loadingFolders > 0;
+  }
+
+  addToSelectedFiles(item: Item, mode: string) {
```
als364: `mode` is never used.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -92,4 +229,77 @@ export class FileDownloadModalComponent  {
...
+
+    if (parent.type === 'File') {
+      return false;
+    }
+
+    if (child === undefined) {
+      return false;
+    }
+
+    // const splitParent = parent.path;
```
als364: Please remove commented out code.

pshapiro4broad:
I'm seeing a case where the UI will eventually freeze, with the Download modal dialog showing. This happens after select/download/cancel, select/download/cancel, which I did a few times, so I'm not sure exactly how to reproduce the problem. I can still bring up the inspector but I'm not sure what to look at to figure out what's going on.

src/app/file-manager/file-manager.module.ts
```diff
@@ -46,7 +49,8 @@ import { GoogleLoginService } from '@app/file-manager/services/login-google.serv
     SharedModule,
     FileManagerRoutingModule,
     StoreModule.forFeature('transferables', TransferablesReducer),
-    MatDialogModule
+    MatDialogModule,
+    // NgxSpinnerModule
```
als364: Please remove commented out code.

src/app/file-manager/models/item.ts
```diff
@@ -41,10 +49,22 @@ export class Item {
...
     this.open = open;
+    this.workspaceName = workspaceName;
+    this.displayName = displayName;
     this.setProgressData();
     this.namespace = namespace;
-    }
+
+    this.children = [];
+    this.childCount = this.children.length;
+    // this.parent = this.prefix.split('/').slice(0, -1).join('/');
```
als364: Please remove commented out code.

src/app/file-manager/file-explorer/file-explorer.component.ts
pshapiro4broad: OK, I see now. Comments would be helpful.

src/app/file-manager/breadcrumb/breadcrumb.component.ts
```diff
@@ -23,30 +28,52 @@ export class BreadcrumbComponent implements OnInit {
...
-    });
+  private processItems(items: Item[]) {
+    this.crumbs = [];
+
+    if (items === null || items === undefined || items.length <= 0) {
+      return;
+    }
+
+    items.forEach(
+      (item, index) => {
```
pshapiro4broad: `index` is unused here, does it need to be declared in the anonymous function argument list?

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -11,38 +10,152 @@ import { DownloadValidatorService } from '@app/file-manager/services/download-va
...
+    return this.loadingFiles || this.loadingFolders > 0;
+  }
+
+  addToSelectedFiles(item: Item, mode: string) {
+    const found: Item = this.selectedFiles.find(x => x.mediaLink === item.mediaLink);
+    if (found === undefined) {
+      this.selectedFiles.push(item);
+      this.fileCount++;
+      this.totalSize += Number(item.size);
+      this.fileCountObservable.next(this.fileCount);
```
pshapiro4broad, line 166: The code looks correct, but when I run it, I don't see the values updating after each API request. E.g. for Workspace6 it just stays at 0 even after waiting for 20 minutes.

Also it updates the observable once for each file. Unless this is very fast, it seems like you'd be better off computing `fileCount` and `totalSize` for the complete API response and then update the observables once per API reply. I would expect that the API request is the slower part of this code and the totaling up of the count and size would be relatively fast. Another approach would be to only update the observables once for each X files processed.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -11,38 +10,152 @@ import { DownloadValidatorService } from '@app/file-manager/services/download-va
...
+        }
+      );
+  }
+
+  isLoading() {
+    return this.loadingFiles || this.loadingFolders > 0;
+  }
+
+  addToSelectedFiles(item: Item, mode: string) {
+    const found: Item = this.selectedFiles.find(x => x.mediaLink === item.mediaLink);
```
pshapiro4broad: Can this lookup be done using a hash map instead of a linear search? Without a hash, this will be O(n^2) for each bucket API call made.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
dags: I think we could. Testing it now ...

src/app/file-manager/file-download-modal/file-download-modal.component.ts
dags: it was used while debugging. Will remove it

src/app/file-manager/file-download-modal/file-download-modal.component.ts
dags: makes sense. Will fix it.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -58,6 +171,7 @@ export class FileDownloadModalComponent  {
   }
 
   cancel(): void {
+    BucketService.delimiter = null;
```
pshapiro4broad: I'm not sure what this is doing. As far as I can tell, `BucketService.delimiter` is unused. If it is needed, this code should call a method on `BucketService` to perform this operation, e.g. `this.bucketService.reset()`.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -74,15 +188,18 @@ export class FileDownloadModalComponent  {
...
-          Type.DOWNLOAD, ItemStatus.PENDING, '', '', '', this.preserveStructure, false, '');
-        this.downloadFiles.push(dataFile);
-        this.store.dispatch(new Transferables.AddItem(dataFile));
-        this.done.emit(true);
-        this.router.navigate(['/status']);
+        if (!this.filesMap.has(file.id)) {
+          this.filesMap.set(file.id, file);
+          const dataFile: Item = new Item(file.id, file.name, file.updated, file.created,
+            file.size, file.mediaLink, file.path, this.directory,
+            Type.DOWNLOAD, ItemStatus.PENDING, '', '', '', this.preserveStructure, false, '', file.displayName, '');
```
pshapiro4broad, line 210: Why do we store `preserveStructure` for each item. Isn't the value the same for all of the items?

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -92,4 +209,68 @@ export class FileDownloadModalComponent  {
...
+    }
+
+    if (child === undefined) {
+      return false;
+    }
+
+    return child.path.startsWith(parent.path);
+  }
+
+  cloneItem(item: Item): Item {
```
pshapiro4broad: It looks like this should be a method on `Item`

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -92,4 +209,68 @@ export class FileDownloadModalComponent  {
...
+
+    const newItem: Item = new Item(
+      item.id,
+      item.name,
+      item.created,
+      item.updated,
+      item.size,
+      item.mediaLink,
+      item.workspaceName + '/',
```
pshapiro4broad: Why add `/` here? Add a comment.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -92,4 +209,68 @@ export class FileDownloadModalComponent  {
...
+    if (child === undefined) {
+      return false;
+    }
+
+    return child.path.startsWith(parent.path);
+  }
+
+  cloneItem(item: Item): Item {
+
+    const newItem: Item = new Item(
```
pshapiro4broad: The variable isn't required here, you can say `return new Item(...);`

src/app/file-manager/file-explorer-upload/file-explorer-upload.component.ts
```diff
@@ -213,9 +213,9 @@ export class FileExplorerUploadComponent implements OnInit {
         this.selectedFiles
           .filter(file => file.data.type === Type.FILE)
           .forEach(file => {
-            this.dataFile =  new Item(UUID.UUID(),  file.data.name, file.data.updated, file.data.updated, file.data.size,
-                          '', file.data.path, result.directory, Type.UPLOAD, ItemStatus.PENDING, '',
-              '', '', result.preserveStructure, false, '');
+            this.dataFile = new Item(UUID.UUID(), file.data.name, file.data.updated, file.data.updated, file.data.size,
+              '', file.data.path, result.directory, Type.UPLOAD, ItemStatus.PENDING, '',
+              '', '', result.preserveStructure, false, '', '', '');
```
pshapiro4broad, line 218: Can there be a factory method in `Item()` that takes care of providing all the missing arguments? It looks like we have many calls to this constructor where a number of the parameters are always the same.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -92,4 +210,74 @@ export class FileDownloadModalComponent  {
...
+
+  /**
+   * This function filters repeated folders when a parent
+   * is all-selected and there are sub-folders in the list,
+   * Items with indeterminate === true must be removed because
+   * selected content is included as other items.
+   * All this extra, non relevante here, items are required
+   * in the selection process.
+   */
+  filterFolderItems(sortedItems): Item[] {
```
pshapiro4broad: Declare types where possible, e.g. here this should be `sortedItems: Item[]`

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -92,4 +210,74 @@ export class FileDownloadModalComponent  {
...
+      if (isDad) {
+        folderItems.splice(i + 1, 1);
+      } else {
+        i++;
+      }
+    }
+    return singleItems;
+  }
+
+  isParent(parent: Item, child: Item): boolean {
```
pshapiro4broad: Could be a method on `Item`, e.g. either `isChildOf()` or `isParentOf()` depending on how you want to call it.

src/app/file-manager/models/bucket-object-db.ts
```diff
@@ -4,6 +4,25 @@ import { BehaviorSubject } from 'rxjs/BehaviorSubject';
 
 export class BucketObjectsDB {
```
pshapiro4broad: This class appears to be unused.

src/app/file-manager/models/item-status.ts
```diff
@@ -4,4 +4,5 @@ export class ItemStatus {
   static readonly UPLOADING = 'Uploading';
   static readonly COMPLETED = 'Completed';
   static readonly CANCELED = 'Canceled';
+  static readonly INDETERMINATE = 'indeterminate';
```
pshapiro4broad: This appears to be unused

src/app/file-manager/models/item.ts
```diff
@@ -19,13 +18,22 @@ export class Item {
   transferred: number;
   remaining: number;
   bucketName: string;
+  workspaceName: string;
   prefix: string;
   delimiter: string;
   open: boolean;
+  selected: boolean;
+  indeterminate: boolean;
```
pshapiro4broad, line 26: Using a enum to represent the three different selection states instead of having two boolean fields could simplify the code, as it would make it easer to reset and read this value in the code. The two booleans can represent four states but only three are used, as far as I can tell. Since the values of each relies on the other, an enum should make the code simpler.

src/app/file-manager/models/item.ts
```diff
@@ -19,13 +18,22 @@ export class Item {
...
   namespace: string;
+  parent?: Item;
+  children?: Item[];
+  childCount?: number;
```
pshapiro4broad: This is set but it doesn't seem to be used.

src/app/file-manager/models/item.ts
```diff
@@ -19,13 +18,22 @@ export class Item {
...
+  removeit = false;
```
pshapiro4broad: left in by mistake?

src/app/file-manager/models/item.ts
```diff
@@ -41,10 +49,21 @@ export class Item {
...
+    this.displayName = displayName;
     this.setProgressData();
     this.namespace = namespace;
-    }
+
+    this.children = [];
+    this.childCount = this.children.length;
+  }
+
+  setChildren(children: Item[]) {
```
pshapiro4broad: This method appears to be unused.

src/app/file-manager/models/selection-status.ts
```diff
@@ -0,0 +1,6 @@
+import { Item } from '../models/item';
+export interface SelectionStatus {
```
pshapiro4broad: This appears to be unused

src/app/file-manager/services/bucket.service.ts
```diff
@@ -2,72 +2,161 @@ import { Injectable } from '@angular/core';
 import { GcsService } from './gcs.service';
 import { Item } from '../models/item';
 import { Observable } from 'rxjs/Observable';
-import { UUID } from 'angular2-uuid';
-import { Type } from '@app/file-manager/models/type';
+import { FirecloudApiService } from '@app/file-manager/services/firecloud-api.service';
```
pshapiro4broad: This appears to be unused

src/app/file-manager/services/bucket.service.ts
```diff
@@ -2,72 +2,161 @@ import { Injectable } from '@angular/core';
...
+
+  }
   readonly DELIMITER = '/';
 
-  public getBucketData(bucketName: String, delimiter: String, token: String, workspaceName: String): Observable<Item[]> {
-    this.elements = [];
+  public getBucketData(parentNode: Item, token: String, useDelimiter: Boolean, processData: boolean): Observable<Item[]> {
     this.isResponseComplete = false;
-    return this.getBucketFiles(bucketName, token, workspaceName, delimiter);
+    return this.getBucketFiles(parentNode, token, useDelimiter, processData);
```
pshapiro4broad: Please add types for all arguments, if possible.

src/app/file-manager/breadcrumb/breadcrumb.component.ts
dags: remains of already deleted code. Will be deleted

src/app/file-manager/file-download-modal/file-download-modal.component.ts
dags: this is an "still evolving" approach trying to update count and sizes as responsive as possible. It is improved in a separated branch waiting for a "bug" story to fix it. I don't think we should address this on current PR.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
dags: looks like more remains of unused code. Will delete.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
dags: Items could  be of different types : Type.DOWNLOAD, Type.UPLOAD, ??, and each type / batch could have a different preserveStructure value. That is the original reason to have it this way. 

We may think another approach but doesn't seem to be a trivial change to be addressed in a PR. We could create a story for that if it is really worth it.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
dags: Right now is a file-download-modal class specific method, makes sense only there, no need to be included in Item class I think.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
dags: Comment added. Due to different data models for FireCloud workspaces and Bucket files, we used this trick to homogeneize  items for downloading preflight

src/app/file-manager/file-download-modal/file-download-modal.component.ts
dags: sure. Another console.log remains.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
dags: done.

src/app/file-manager/models/item-status.ts
dags: Yes, it was a mix of concepts  so it was moved to another property. Will remove it.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
dags: Ok. Is used only there but will add it to Item.

src/app/file-manager/models/bucket-object-db.ts
dags: ok, deleted.

src/app/file-manager/models/item.ts
dags: selected and indeterminate are two different properties. Each value is handled by a different checkbox property as well. Not sure to understand this comment. Changing this implies a lot of changes in the code and will require to go through QA again.  All design was done with two different properties in mind.

src/app/file-manager/models/item.ts
dags: children is used, parent and childCount don't so will be removed.

src/app/file-manager/models/item.ts
dags: removed

src/app/file-manager/models/item.ts
dags: removed

src/app/file-manager/models/selection-status.ts
dags: removed

src/app/file-manager/services/bucket.service.ts
dags: FirecloudApiService is used in several places.

src/app/file-manager/services/bucket.service.ts
dags: ??.  I guess this is an error. Types can't be added here.

dags:
Last PR fixes break the app. It is not working. I'll review it and let you know when is working again.

dags:
App is fixed and working as expected again.

src/app/file-manager/file-explorer-upload/file-explorer-upload.component.ts
dags: I reviewed this. There just few calls to constructor and I don't see a big win in introducing the factory method.

src/app/file-manager/models/item.ts
pshapiro4broad: OK, maybe I don't understand what these fields mean. I thought 'selected' is true when the item is selected, and 'indeterminate' is true when the item is partially selected (e.g. it's a folder with only some of its contents selected). They are both false when the item is neither selected or partially selected. Therefore there is no use case where they are both true, and we only have three states. Looking at the code, it looked like they are almost always both updated at the same time, another sign that they are representing a single (composite) value. If this is all true, than creating an enum with three states would be a valid replacement.

If this seems to much work to do now, create a JIRA for the work so we can address it later.

src/app/file-manager/services/bucket.service.ts
pshapiro4broad: Where is it used in this file? The only occurrence of `FirecloudApiService` in this file that I see is on this import line.

src/app/file-manager/services/bucket.service.ts
pshapiro4broad: Is there no way to add types to the `getBucketFiles()` method on line 25?

pshapiro4broad (approved):
Looks good, just a few more minor comments

src/app/file-manager/services/bucket.service.ts
dags: got it. I miss-understood your comment.

src/app/file-manager/services/bucket.service.ts
dags: ok, I was looking only to line 24. Line 25 is not shown here ... My fault ...

PR #63, opened 2018-03-20
lf-DSDEGP-2179-Implement 'Cancel export' from bucket to bucket

LeoForconesi:
### Description

_This code completes bucket to bucket export, adding the ability to cancel all exports in 'Pending' state._

_This work is related to [DSDEGP-2179](https://broadinstitute.atlassian.net/browse/DSDEGP-2179). in Jira's board_

----

### Checklist

* Review cycle:


src/app/file-manager/services/gcs-api.service.ts
```diff
@@ -61,18 +61,18 @@ export class GcsApiService extends GcsService {
   public cancelAll() {
     const itemsToDownload = this.getDownloadingFiles();
     const itemsToUpload = this.getUploadingFiles();
-    const itemsToExport = this.getExportPendingFiles();
+    const itemsToExport = this.getExportingFiles();
     if (itemsToDownload.length > 0) {
```
pshapiro4broad: Does this code need to check this here? The same length check is done in `cancelDownloads()` below. Same is true for the other cancels.

src/app/file-manager/services/gcs-api.service.ts
```diff
@@ -92,11 +92,11 @@ export class GcsApiService extends GcsService {
     }
   }
 
-  public cancelExportsToGCP(): Observable<boolean> {
-    const items = this.getExportPendingFiles();
+  public cancelExportsToGCP() {
+    this.cancel = true;
```
pshapiro4broad: It looks like this is only used for GCP, if so, its name should reflect that.

src/app/file-manager/services/gcs-api.service.ts
```diff
@@ -133,33 +133,21 @@ export class GcsApiService extends GcsService {
   }
 
   getDownloadingFiles() {
-    const items = new FilesDatabase(this.store).data.
-      filter(item => item.status === ItemStatus.DOWNLOADING ||
-        (item.status === ItemStatus.PENDING && item.type === Type.DOWNLOAD));
+    const items = new FilesDatabase(this.store).data.filter(item => item.status === ItemStatus.DOWNLOADING ||
+      (item.status === ItemStatus.PENDING && item.type === Type.DOWNLOAD));
```
pshapiro4broad: It looks like these three methods could be refactored into a single method that accepts a `Type` parameter. There already is `getFiles()` which is pretty much that method, minus the refactoring.

Also, there's no need for an intermediate variable `items` here.

src/app/file-manager/services/gcs-api.service.ts
```diff
@@ -92,11 +92,11 @@ export class GcsApiService extends GcsService {
     }
   }
 
-  public cancelExportsToGCP(): Observable<boolean> {
-    const items = this.getExportPendingFiles();
+  public cancelExportsToGCP() {
+    this.cancel = true;
+    const items = this.getExportingFiles();
     if (items.length > 0) {
-      this.openModal('export-gcp-cancel', 'cancelAllExportsToGCP', Type.EXPORT_GCP);
```
pshapiro4broad: Why doesn't this open a modal dialog and send the `export-gcp-cancel` message?

src/app/file-manager/services/gcs-api.service.ts
```diff
@@ -61,18 +61,18 @@ export class GcsApiService extends GcsService {
...
-      this.cancelItemsStatus(itemsToDownload);
+      this.cancelItemsStatus(Type.DOWNLOAD);
     }
     if (itemsToUpload.length > 0) {
       this.electronService.ipcRenderer.send('upload-cancel');
-      this.cancelItemsStatus(itemsToUpload);
+      this.cancelItemsStatus(Type.UPLOAD);
     }
     if (itemsToExport.length > 0) {
       this.electronService.ipcRenderer.send('export-gcp-cancel');
```
pshapiro4broad: This is being sent but I don't see where a handler is registered to receive it.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
```diff
@@ -126,21 +130,32 @@ export class TransferablesGridComponent implements OnInit, AfterViewInit {
...
-        this.modeVariable = this.DETERMINATE;
-        this.generalExportToGCPProgress = 100;
-        TransferablesGridComponent.isExporting = false;
-      } else if (TransferablesGridComponent.isExporting) {
-        this.modeVariable = this.INDETERMINATE;
-        this.limitTransferables.exportItems(items);
-      }
-    });
+
+    this.gcsService.exportItemCompleted.subscribe(() => {
```
pshapiro4broad, line 133: This seems like an odd way for an API user to monitor results from a method call. I would expect the notification method to be passed as a parameter to `exportToGCPFiles()`. The subscribe can be done inside the API implementation, so you don't need to make `exportItemCompleted` part of the abstract API.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
```diff
@@ -26,6 +27,7 @@ import { Observable } from 'rxjs/Observable';
 export class TransferablesGridComponent implements OnInit, AfterViewInit {
 
   static isExporting: Boolean = false;
```
pshapiro4broad: Do these fields have to be static? Can't any object that needs their values get them from the component? We shouldn't need to use global state to store these values.

pshapiro4broad (commented):
A few comments, nothing minor

pshapiro4broad:
Issues encountered when testing:

- need to prompt at exit for cancel operation
- using indeterminate progress for overall export progress; since total number of files and current export # of files is known, this can use a determinate progress bar
- performance issues when exporting many files -- using a test case of 24000 files, performance seems very slow

wloforteBroad:
About issues:
- First two where discussed earlier and Leo is working on that.
- Performance issues are known and will be addressed in different User story.

pshapiro4broad:
I don't see the change in this PR, but I think the text `Exports To GCP` is not the best way to describe this in the UI. I don't think most users will know what GCP is. I think `export to google cloud` (or `export to cloud`, which matches the current UI, would be better.

src/app/file-manager/services/gcs-api.service.ts
LeoForconesi: Change made will only check store's items only once when cancelAll is called.

src/app/file-manager/services/gcs-api.service.ts
LeoForconesi: Changed to a more logic name

src/app/file-manager/services/gcs-api.service.ts
LeoForconesi: You are right! this is now refactored

src/app/file-manager/services/gcs-api.service.ts
LeoForconesi: This is being handled in transferables-grid.component in order to handle cancel action in a more directly way.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
LeoForconesi: You're right in this observation, is it ok for you if we create a Jira story for this fix? This approach is introducing some buggy behavior in the UI that we need to handle carefully.

LeoForconesi:
- Prompt at exit for cancel operation is now working.
- General progress is now determinate for exports progress.
- Changed wording for 'Exports To GCP' to 'Export to google cloud'

src/app/file-manager/services/gcs-api.service.ts
LeoForconesi: This has been removed.

src/app/file-manager/services/gcs-api.service.ts
```diff
@@ -59,44 +54,39 @@ export class GcsApiService extends GcsService {
...
-      this.cancelItemsStatus(itemsToDownload);
-    }
-    if (itemsToUpload.length > 0) {
-      this.electronService.ipcRenderer.send('upload-cancel');
-      this.cancelItemsStatus(itemsToUpload);
-    }
-    if (itemsToExport.length > 0) {
-      this.electronService.ipcRenderer.send('export-gcp-cancel');
-      this.cancelItemsStatus(itemsToExport);
+    if (new FilesDatabase(this.store).data.filter(item => item.status === ItemStatus.PENDING
```
pshapiro4broad: I don't think this logic is correct, or it's at least redundant, as it mentions `PENDING` twice.

I know that `&&` takes precedence over `||` but I find that it's clearer to use `(` `)` around the `&&` expression so there's no confusion, assuming that you need to keep the `&&` part of this expression.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
pshapiro4broad: OK

electron_app/helpers/enviroment.js
```diff
@@ -13,8 +13,8 @@ const constants ={
   GOOGLE_API: 'https://www.googleapis.com/upload/storage/v1/b/',
   IPC_VERIFY_BEFORE_DOWNLOAD: 'verify-before-download',
   IPC_UPLOAD_STATUS: 'upload-status',
-  IPC_UPLOAD_CANCEL: 'upload-cancel',
-  IPC_DOWNLOAD_CANCEL: 'download-cancel',
+  IPC_UPLOAD_CANCEL: 'upload-cancelGCPExports',
```
pshapiro4broad, line 16: ~~It looks like `IPC_UPLOAD_CANCEL` and `IPC_DOWNLOAD_CANCEL` are currently unused. The idea was that we would use a constant instead of having the message name in our code, to avoid duplicating magic strings. But right now nothing is using these, apparently. So either we should update the code to use them, or remove them from here.~~

~~Looking at the code more, I don't see any code that registers an `on()` handler for these messages, so I don't think the `send()`s are having an effect. If not, they should be removed, and the unused constants should be removed too.~~

Never mind, for some reason IntelliJ is refusing to include `main.js` in the project's source files.

src/app/file-manager/services/gcs-api.service.ts
```diff
@@ -59,44 +54,39 @@ export class GcsApiService extends GcsService {
...
+      this.cancelItemsStatus(Type.EXPORT_GCP);
     }
   }
 
   public cancelDownloads(): Promise<boolean> {
-    const items = this.getDownloadingFiles();
-    if (items.length > 0) {
-      this.openModal('download-cancel', 'cancelAllDownloads', Type.DOWNLOAD);
+    if (this.getFiles(Type.DOWNLOAD).length > 0) {
+      this.openModal('download-cancelGCPExports', 'cancelAllDownloads', Type.DOWNLOAD);
```
pshapiro4broad: I'm not sure why this event name was changed. Does it now do something different?

src/app/file-manager/services/gcs-api.service.ts
LeoForconesi: This is now fixed

electron_app/helpers/enviroment.js
LeoForconesi: What you said is right, this variable was not used on send from angular's renderer. This is now used as it should. In the next PR we will double check that all constants are being properly used.

pshapiro4broad (approved):

PR #64, opened 2018-03-23
DSDEGP - 2103 - Implement UploadToAmazon infrastructure

LeoForconesi:
### Description

_In a first iteration for exporting Google Buckets to Amazon buckets, we are downloading the requestesd file to the local machine and then performing the export to amazon. In a second iteration we will use Streaming for this feature._

_Jira Tickets included in this PR:_
[DSDEGP-2102](https://broadinstitute.atlassian.net/browse/DSDEGP-2102) : UI implementation
[DSDEGP-2103](https://broadinstitute.atlassian.net/browse/DSDEGP-2103) : APIS Implementation
[DSDEGP-2106](https://broadinstitute.atlassian.net/browse/DSDEGP-2106) : Local Disk handling
[DSDEGP-2107](https://broadinstitute.atlassian.net/browse/DSDEGP-2107) : Tracking transfers

----

### Checklist

* Review cycle:


electron_app/DownloadManager.js
```diff
@@ -62,10 +62,13 @@ const destroyDownloads = () => {
...
 const setHeader = (access_token) => {
   return {
-    count: 8,
+    count: 1,
     port: 443,
     headers: {
       'Authorization': 'Bearer ' + access_token,
+    },
+    onStart: function (meta) {
+      console.log('META', meta);
```
pshapiro4broad: Test code like this should be conditional on `environment.testing` or some similar construct

electron_app/DownloadManager.js
```diff
@@ -62,10 +62,13 @@ const destroyDownloads = () => {
 
 const setHeader = (access_token) => {
   return {
-    count: 8,
+    count: 1,
```
pshapiro4broad: What does this change do?

electron_app/Downloader.js
```diff
@@ -23,7 +23,7 @@ const Downloader = function () {
 
 Downloader.prototype._defaultOptions = {
   //To set the total number of download threads
-  threadsCount: 2, //(Default: 6)
+  threadsCount: 1, //(Default: 6)
```
pshapiro4broad: Does this change make all google cloud downloads single threaded? If so, does it affect performance of downloads outside of the export to S3 feature?

electron_app/ExportS3.js
```diff
@@ -0,0 +1,105 @@
...
+const uploadS3 = (data) => {
+  const filePathExport = data.preserveStructure ? data.item.path : data.item.displayName;
+  const tempFilePath = path.join(tempPath, 'TempFolder', data.item.displayName);
+  const fileStream = fs.createReadStream(tempFilePath);
+  const putParams = {
+    Bucket: data.bucketName,
+    Key: 'Imports/' + filePathExport,
+    Body: fileStream
+  };
+  fileStream.on('error', (error) => { console.log(error) });
```
pshapiro4broad, line 76: Do we need better error handling in this code? Am not sure what errors show up via the error handlers here. Maybe make a JIRA to look at this later, unless we're sure it's not necessary.

electron_app/ExportS3.js
```diff
@@ -0,0 +1,105 @@
...
+  let s3 = new AWS.S3();
+  s3.putObject(putParams, (putErr, putData) => {
+    if (putErr) console.log(putErr);
+    else {
+      removeFile(tempFilePath, data.item);
```
pshapiro4broad: It seems like we'd need a better way to clean up temporary files, in the case where an error occurs or if the transfer is canceled. But maybe that's not worth doing if we're replacing it with the streaming APIs soon.

electron_app/ExportS3.js
```diff
@@ -0,0 +1,105 @@
...
+  s3List.push(s3);
+};
+
+const removeFile = (filePath, item) => {
+  fs.unlink(filePath, (err) => {
+    if (err)
+      console.log('Error removing file ', err);
+    tempPath = pathTemp;
+    item.status = 'Exported to S3';
+    electronWin.webContents.send(constants.IPC_EXPORT_S3_DOWNLOAD, item);
```
pshapiro4broad: maybe this message should have `STATUS` in is name?

electron_app/helpers/handleDisk.js
```diff
@@ -1,8 +1,9 @@
 const fs = require('fs');
 const mkdirp = require('mkdirp');
 const diskspace = require('diskspace');
-
-let saveConditions = { enoughSpace: Boolean, allowed: Boolean };
+const path = require('path');
```
pshapiro4broad: `path` appears to be unused

src/app/app.component.ts
```diff
@@ -63,7 +61,8 @@ export class AppComponent implements OnInit, OnDestroy {
     filter(item =>
          item.status === ItemStatus.DOWNLOADING
       || item.status === ItemStatus.UPLOADING
-      || (item.status === ItemStatus.PENDING && Type.EXPORT_GCP));
+      || (item.status === ItemStatus.PENDING && Type.EXPORT_GCP
+      || item.status === ItemStatus.EXPORTING_S3));
```
pshapiro4broad: I think the `Type.EXPORT_GCP` is incorrect (from before), shouldn't it be `(item.status === PENDING && item.type == Type.EXPORT_GCP)`. Also it looks like the parens are around the wrong subexpression.

src/app/file-manager/file-explorer/file-explorer.component.html
```diff
@@ -81,6 +81,15 @@ <h1>Select files to
     </div>
   </div>
 
+  <div *ngIf="exportInProgress" class="warn-message">
```
pshapiro4broad: There should be a way to avoid the duplicate markup here. But that can wait for a future PR.

src/app/file-manager/services/electron-ipc.api.service.ts
```diff
@@ -38,5 +40,35 @@ export class ElectronIpcApiService extends ElectronIpcService {
   public logout(): void {
     this.electronService.ipcRenderer.send(constants.IPC_GOOGLE_LOGOUT);
   }
-}
 
+  public exportS3(dataTransfer: any): void {
+    this.electronService.ipcRenderer.removeAllListeners(constants.IPC_EXPORT_S3);
```
pshapiro4broad: I think this came up before, did we try using `once()` to avoid having to use `removeAllListeners()`?

src/app/file-manager/services/gcs-api.service.ts
```diff
@@ -46,28 +46,35 @@ export class GcsApiService extends GcsService {
...
+        || item.status === ItemStatus.EXPORTING_S3).length > 0) {
 
       this.electronService.ipcRenderer.send(constants.IPC_DOWNLOAD_CANCEL);
-      // IPC_DOWNLOAD_CANCEL: 'download-cancelGCPExports',
       this.cancelItemsStatus(Type.DOWNLOAD);
 
       this.electronService.ipcRenderer.send(constants.IPC_UPLOAD_CANCEL);
       this.cancelItemsStatus(Type.UPLOAD);
 
+      this.electronService.ipcRenderer.send('download-cancel');
```
pshapiro4broad: Is this correct? We are already sending `IPC_DOWNLOAD_CANCEL` above.

src/app/file-manager/services/gcs-api.service.ts
```diff
@@ -93,6 +100,11 @@ export class GcsApiService extends GcsService {
     }
   }
 
+  public cancelExportToS3() {
+    this.electronService.ipcRenderer.send('download-cancel');
```
pshapiro4broad: This doesn't make sense to me; don't we want to do this only after the dialog is shown?

src/app/file-manager/services/limit-transferables.service.ts
```diff
@@ -50,14 +51,20 @@ export class LimitTransferablesService {
...
+        this.store.dispatch(new Transferables.UpdateItem(item));
+      }
     });
 
     if (type === Type.DOWNLOAD) {
       this.gcsService.downloadFiles(maxFiles);
     } else if (type === Type.UPLOAD) {
       this.gcsService.uploadFiles(localStorage.getItem('uploadBucket'), maxFiles);
+    } else if (type === Type.EXPORT_S3) {
+      maxFiles[0].status = ItemStatus.EXPORTING_S3;
```
pshapiro4broad: Not sure why this code is using `maxFiles[0]`, do we know that there can only be one? A comment would be helpful here.

src/app/file-manager/services/s3-export.service.ts
```diff
@@ -0,0 +1,47 @@
...
+      'accessKey': SecurityService.getS3AccessKey(),
+      'secretKey': SecurityService.getS3SecretKey(),
+      'bucketName': localStorage.getItem('S3BucketName'),
+    };
+    this.electronService.setCredentials(dataTransfer);
+  }
+
+  public startUpload(item) {
+    const dataTransfer = {
+      'accessKey': SecurityService.getS3AccessKey(),
```
pshapiro4broad: Do we need to send AWS credentials both on `test` and on `start upload` methods? I would think once you're authenticated with AWS you wouldn't need to do it again.

src/app/file-manager/transferables-grid/transferables-grid.component.html
```diff
@@ -81,9 +94,9 @@ <h4 class="subtitle">Exported: {{filesDatabase.exportingCount + ' of ' + filesDa
         <mat-cell *matCellDef="let row">
           <mat-icon class="transfer-icon" *ngIf="row.type == 'Download'">file_download</mat-icon>
           <mat-icon class="transfer-icon" *ngIf="row.type == 'Upload'">file_upload</mat-icon>
-          <i class="export-icon-dark" *ngIf="row.type == 'Export GCP'"></i>
+          <i class="export-icon-dark" *ngIf="row.type == 'Export GCP' || row.type == 'Export S3' "></i>
           {{row.status}}
-          <div *ngIf="row.status === 'Downloading' || row.status === 'Uploading' || row.status === 'Exporting to GCP'" class="progress-dots transfer-status">
+          <div *ngIf="row.status === 'Downloading' || row.status === 'Uploading' || row.status === 'Exporting to Google Cloud' || row.status === 'Exporting to Amazon'" class="progress-dots transfer-status">
```
pshapiro4broad, line 99: Relying on text like this seems somewhat fragile, as the wording might change at some point. Is there a better way to do this, e.g. using a fixed token in the row that's not visible in the UI to control this?

pshapiro4broad (commented):
A few mostly minor comments

electron_app/DownloadManager.js
tadeork: I'll fix now. I just was testing something related to the streaming before starting with the download file process. It shouldn't be here.

electron_app/Downloader.js
tadeork: This is also related to that test.

electron_app/ExportS3.js
tadeork: I think we should leave this due to we're going to make the transfer through streaming.

src/app/file-manager/services/electron-ipc.api.service.ts
tadeork: Even more, I think we don't need to remove listeners here.

electron_app/ExportS3.js
tadeork: This will not be necessary when we switch to streaming. But it may be helpful to have a ticket in JIRA to look at this later.

pshapiro4broad: OK, something to keep in mind when working on streaming, then.

src/app/file-manager/transferables-grid/transferables-grid.component.html
tadeork: I think the best for this is create a separate ticket.

pshapiro4broad (approved):

PR #67, opened 2018-04-04
DSDEGP-2151 & 2238: Export progress bar fix & Overall download progress display doesn't remove items that are transferred

LeoForconesi:
### Description

_All filters used to count and display progress have been modified in order to reflect real progress._
_Also some modifications are done here for Overall download progress to display items that are transferred._

_**Jira Stories included in this Code:**_

- [DSDEGP-2151](https://broadinstitute.atlassian.net/browse/DSDEGP-2151) : Overall download progress display doesn't remove items that are transferred

- [DSDEGP-2238](https://broadinstitute.atlassian.net/browse/DSDEGP-2238) : Export Progress bar not working as expected

----

### Checklist

- [] **Submitter**: Ensure that you have added or updated tests
* Review cycle:


src/app/file-manager/reducers/transferables.reducer.ts
```diff
@@ -8,111 +8,114 @@ import { Type } from '@app/file-manager/models/type';
 export type Action = TransferablesActions.All;
 
 export class TransferableState {
-  items: Item[];
+    items: Item[];
```
pshapiro4broad: Looks like a re-indent issue; I believe we're using 2 space indent for these files, not 4 space.

LeoForconesi: indeed, we are using 2 space indent, this is now fixed.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -12,6 +12,12 @@ import { Router } from '@angular/router';
 import { ItemStatus } from '@app/file-manager/models/item-status';
 import { TransferablesGridComponent } from '@app/file-manager/transferables-grid/transferables-grid.component';
 import { PreflightService } from '../services/preflight.service';
+import { FilterSizePipe } from '../filters/filesize-filter';
```
pshapiro4broad: remove unused imports, `FilterSizePipe`, `Observable`, `BehaviorSubject`

src/app/file-manager/file-download-modal/file-download-modal.component.ts
```diff
@@ -135,4 +143,13 @@ export class FileDownloadModalComponent implements OnInit {
   selectedFiles() {
     return this.preflightService.selectedFiles;
   }
+
+  updateCurrentBatch() {
+    const items = new FilesDatabase(this.store).data.filter(item => item.type === Type.DOWNLOAD && item.currentBatch);
```
pshapiro4broad: Is there a way to avoid the code duplication here, such as by making this method accept `type`?

src/app/file-manager/models/item.ts
```diff
@@ -26,10 +26,11 @@ export class Item {
   indeterminate: boolean;
   namespace: string;
   children?: Item[];
+  currentBatch: boolean;
 
   constructor(id, name, created, updated, size, mediaLink, path, destination,
     type, status, bucketName, prefix, delimiter, preserveStructure, open,
-    workspaceName, displayName, namespace) {
+    workspaceName, displayName, namespace, currentBatch) {
```
pshapiro4broad: These arguments should all have declared types, e.g. `currentBatch: boolean`

src/app/file-manager/services/gcs-api.service.ts
```diff
@@ -160,7 +160,7 @@ export class GcsApiService extends GcsService {
         'Authorization': 'Bearer ' + SecurityService.getAccessToken()
       })
     };
-    return this.http.post(url, null, httpOptions);
+    return this.http.post(url, null, httpOptions).retry(1);
```
pshapiro4broad: Why is this change necessary? A comment would be helpful.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
```diff
@@ -28,10 +28,12 @@ import { WarningModalComponent } from '../warning-modal/warning-modal.component'
 
 export class TransferablesGridComponent implements OnInit, AfterViewInit {
   static isExporting: Boolean = false;
+  static firstIteration: Boolean = true;
   @ViewChild(MatPaginator) paginator: MatPaginator;
   @ViewChild(MatSort) sort: MatSort;
-  displayedColumns = ['name', 'size', 'status', 'progress', 'actions'];
+  @ViewChild('pbg') pbg: ProgressBar;
```
pshapiro4broad: This appears to be unused; if so, it (and the copy in HTML) should be removed.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
```diff
@@ -28,10 +28,12 @@ import { WarningModalComponent } from '../warning-modal/warning-modal.component'
 
 export class TransferablesGridComponent implements OnInit, AfterViewInit {
   static isExporting: Boolean = false;
+  static firstIteration: Boolean = true;
```
pshapiro4broad: Why does this need to be `static`? It looks like it and `isExporting` can be converted to fields.

pshapiro4broad (commented):
A few minor comments. When I tried this out, I did notice this behavior:

- start a long download, cancel it 
- at this point the Downloaded progress bar is partly full, depending on how much data has been downloaded
- switch to the Download view, then switch back to Status
- the progress bar is now empty

This isn't a major issue, but I was expecting that the progress bar and the "x out of y files" message would be updated when a cancel occurs, since this status no longer applies to the current download. I was expecting that bar would be empty, and the message would say "0 out of 0 files".

src/app/file-manager/services/gcs-api.service.ts
LeoForconesi: I added this change in order to minimize impact from the Out of resources issue that we are experiencing from Chromium limitations when exporting a large amount of files between Google Buckets. 

By default angular's http client will retry multiple times before dropping this request. Meanwhile, the app will continue sending more posts from other exports (these might fail too), causing a request memory overflow, leading to chromium's error.

The idea here is to retry only once if the request fails, stopping any request flood.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
LeoForconesi: The reason these are not local variables, is because their values are not persisting through the app execution.

When receiving info from observables we get a new instance from the component, which redefines all these variables, needed to reflect changes in the Ui. 

We were discussing a possible solution that is implement Redux store in order to reflect the app state allowing us to remove every static variable, but at the moment this would be a temporal solution.

src/app/file-manager/file-download-modal/file-download-modal.component.ts
LeoForconesi: Unused imports were removed.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
LeoForconesi: Sorry about that, this variable is removed.

LeoForconesi:
@pshapiro4broad The behavior you saw with downloads is registered in Jira, and will be solved soon. Thanks for the heads up!

src/app/file-manager/transferables-grid/transferables-grid.component.ts
pshapiro4broad: It's OK to leave these as statics for now. Maybe add a comment about how a new component is created each time but this state has a longer lifetime than the component. The goal would be to put the properties on an object that has the correct lifetime, maybe the Store is the best place.

src/app/file-manager/services/gcs-api.service.ts
pshapiro4broad: That's good, can you add a comment that describes this? Reading the code I had assumed that, by default, no retries would occur and the change was to request a retry.

pshapiro4broad (approved):

PR #68, opened 2018-04-05
DSDEGP 2207 Use streaming APIs for AWS export

tadeork:
### Description

_The export process now is made through streaming via the users computer._
_Now there's no need to download and then upload the files. With this change the export process should be more fast._
_[Use streaming APIs for AWS export](https://broadinstitute.atlassian.net/browse/DSDEGP-2207) Jira ticket._

----

### Checklist

* Review cycle:


electron_app/ExportS3.js
```diff
@@ -1,35 +1,33 @@
...
+  uploadS3(data);
 };
 
-const testCredentials = (data) => {
+const testS3Credentials = (data) => {
   let errorMessage = null;
+  AWS.config.update({ accessKeyId: data.accessKey, secretAccessKey: data.secretKey });
+  
   s3 = new AWS.S3({
     accessKeyId: data.accessKey,
```
pshapiro4broad: Does this need to be provided here as well as on the `AWS.config` call above?

electron_app/ExportS3.js
```diff
@@ -58,44 +56,50 @@ const testCredentials = (data) => {
...
-    if (putErr) console.log(putErr);
-    else {
-      removeFile(tempFilePath, data.item);
-    }
+    Key: 'Imports/' + filePathExport
+  });
+
+  request.get(url, setHeader(data.gcsToken))
+  .on('error', (err) => {
+    console.error(err);
```
pshapiro4broad, line 69: We will need better error handling at some point, since logging to the console isn't visible to the user. If we don't have a JIRA for this yet, please create one.

src/app/file-manager/services/gcs-api.service.ts
```diff
@@ -134,8 +135,7 @@ export class GcsApiService extends GcsService {
         currentStatus = ItemStatus.DOWNLOADING;
         break;
       case Type.EXPORT_S3:
-        currentStatus = ItemStatus.EXPORTING_S3;
-        break;
+        return new FilesDatabase(this.store).data.filter(item => (item.status === ItemStatus.PENDING && item.type === Type.EXPORT_S3));
```
pshapiro4broad: Should add a comment describing why this case is different from the others.

pshapiro4broad:
I ran into an issue when testing this. I started an S3 export for a large file, and then canceled the transfer. The Status page was correctly updated to show that all transfers were canceled but the item was still showing transfer progress. If I switch back to the Export tab, the UI wouldn't let me start another transfer and said that a transfer was still in progress. So it looks like a long running transfer can't be canceled under some circumstances. I was still able to quit the app to force the transfer to stop.

pshapiro4broad (commented):
A few minor comments. However I did see two problems when testing:

- cancel S3 export doesn't seem to work properly
- S3 export on a 0 byte file (workspace3: `/Uploads/Untitled Document 1`) hangs

It's possible that these issues aren't specific to the new streaming code, (e.g. they may exist on `develop` too). If so, don't hold up the merge to fix them, create new JIRAs to track these issues.

electron_app/ExportS3.js
tadeork: Yes, we just made [one](https://broadinstitute.atlassian.net/browse/DSDEGP-2259).

pshapiro4broad (approved):
👍 with the knowledge that we'll have (or already have) JIRAs for the issues I ran into when testing. Even if we choose not to fix them, the issues will help in creating release notes of known issues.

PR #70, opened 2018-04-18
Dg dsdegp 2221

dags:
### Description

Major refactoring to improve performance and avoid freezing under heavy load. 
This refactoring was needed to make application more stable before next release, taking into account that coding might be put on hold for some weeks / months while waitnig for users' feedback.

GAWB-3538 https://broadinstitute.atlassian.net/browse/GAWB-3538
GAWB-3412 https://broadinstitute.atlassian.net/browse/GAWB-3412
GAWB-3534 https://broadinstitute.atlassian.net/browse/GAWB-3534

More information on related stories. 
----

### Checklist

* Review cycle:
- [X] **Submitter**: Verify all tests go green, including CI tests


pshapiro4broad:
Some issues seen when running the client:

- on MacOS, menus other than `Edit` are not included. This is a regression vs `develop`, where there are five menus in addition to the apple menu: Electron, Edit, View, Window, Help.

- the app no longer shows pending or completed items. I know that some users had asked for a way to see what files were downloaded once a download was complete. Was this change intentional? I would expect pending/completed/canceled items to be cleared when the next operation begins.

dags:
Mac menus were not supposed to exist, like occurs in Windows and Linux. So we removed, but after testing we find out that Copy & Paste didn't work if Edit menu is not enabled. That is why Edit menu is the only currently present. Copy & Paste feature is required to paste AWS & Google buckets info and auth keys. 

Completed and Pending items were removed from progress lists because in our tests the application became unstable, almost not working, when exporting thousands of files and root cause was those big progress tables. We discussed this approach with Moran and Ilyana and they agreed that having thousands of Complete and Pending entries in lists doesn't add much value. We'll continue thinking on a better way to handle those huge loads. 

The application will be release quite soon, probably in a week or so and then will enter a feedback period of a few months.

pshapiro4broad:
> Mac menus were not supposed to exist, like occurs in Windows and Linux. So we removed, but after testing we find out that Copy & Paste didn't work if Edit menu is not enabled. That is why Edit menu is the only currently present. Copy & Paste feature is required to paste AWS & Google buckets info and auth keys.

I don't know about Windows and Linux, but on the Mac an app should at least have a App menu in addition to the Edit menu. A minimal App menu contains About, the three Hide menu items, and Quit.

electron_app/UploadManager.js
```diff
@@ -44,28 +48,33 @@ const uploadManager = (bucketName, fileList = [], access_token, win) => {
...
-
         }
-        progressConf.on('progress', function (progress) {
-          file.progress = Math.round(progress.percentage);
-          file.transferred = progress.transferred;
-          win.webContents.send(constants.IPC_UPLOAD_STATUS, file);
+
+        progressConf.on('progress', (progress) => {
+          if (progress.percentage >= 100.0 && progress.delta === 0) {
+            file.progress = 100;
```
pshapiro4broad, line 58: There is a case where % progress is > 100?

dags: Don't know, Just defensive coding against rounding or similar issues.

pshapiro4broad:
In my testing it doesn't look like GAWB-3533 has been addressed. This is the issue where the preflight text isn't updated as more preflight data is generated. To test this out, I selected Workspace6 and then clicked Download selection. After waiting several minutes, the text in the dialog hasn't been updated. I would expect this text to be updated as soon as the first server response comes back.

Also I see that the Cancel button isn't enabled. This seems like a bug. Without enabling Cancel during preflight, a user will be forced to use the OS's task manager to kill the app so they can stop it from doing the preflight work.

wloforteBroad:
That is a mistake, sorry for that. That story is not on review. I moved it back to testing.

src/app/file-manager/models/entity-status.ts
```diff
@@ -0,0 +1,13 @@
+export class EntityStatus {
+
+  static readonly PENDING = 0;
+  static readonly INPROGRESS = 1;
+  static readonly PAUSED = 2;
+  static readonly COMPLETED = 3;
+  static readonly CANCELED = 4;
+  static readonly FAILED = 5;
+
+  static lenght = 6;
```
pshapiro4broad: Typo `length`? It doesn't appear to be used anywhere, can it be removed?

src/app/file-manager/models/export-to-gcs-item.ts
```diff
@@ -0,0 +1,59 @@
...
+    this.delimiter = delimiter;
+    this.selected = false;
+    this.indeterminate = false;
+    this.workspaceName = workspaceName;
+    this.displayName = displayName;
+    this.setProgressData();
+    this.namespace = namespace;
+  }
+
+  setProgressData() {
```
pshapiro4broad, line 52: A more OO solution would put common items into a shared class, or a common base class, to avoid duplication.

src/app/file-manager/models/entity-status.ts
dags: Yes, it is a typo and will be removed.

src/app/file-manager/models/export-to-gcs-item.ts
dags: Do you mean  in xxxx-item.ts classes ?- We had a single item class for all, and later I created a class for each item: download, upload, export to gcs and export to s3, because data needs were different and we saw the need to split reducers to improve performance. We have not finished yet the cleaning of items properties after the refactoring, we expected to continue it in some moment after release. You will find similar things in other classes as well. We tried to keep features as decoupled as possible until we understand well use cases and data needs for each case. We migth create a story to refactor code as you suggest.

src/app/file-manager/models/item-status.ts
```diff
@@ -8,4 +8,18 @@ export class ItemStatus {
   static readonly EXPORTED_S3 = 'Exported to S3';
   static readonly EXPORTING_GCP = 'Exporting to Google Cloud';
   static readonly FAILED = 'Failed';
+
+  static readonly IPENDING = 0;
```
pshapiro4broad, line 12: These don't appear to be used.

src/app/file-manager/models/item-status.ts
```diff
@@ -8,4 +8,18 @@ export class ItemStatus {
   static readonly EXPORTED_S3 = 'Exported to S3';
```
pshapiro4broad, line 8: It looks like `EXPORTED_S3` and `FAILED` are also unused.

src/app/file-manager/models/type.ts
```diff
@@ -5,4 +5,13 @@ export class Type {
      static readonly FOLDER = 'Folder';
      static readonly EXPORT_S3 = 'Export S3';
      static readonly EXPORT_GCP = 'Export GCP';
+
+    static readonly IDOWNLOAD = 0;
+    static readonly IUPLOAD = 1;
```
pshapiro4broad: Some unused members here as well, also indent doesn't appear to be correct.

src/app/file-manager/models/upload-item.ts
```diff
@@ -0,0 +1,149 @@
...
+    this.remaining = 0;
+    this.progress = 0;
+  }
+
+}
+
+
+
+
+// import { Type } from './type';
```
pshapiro4broad: Commented out code should be removed

src/app/file-manager/models/export-to-gcs-item.ts
pshapiro4broad: OK

src/app/file-manager/progress-tab/progress-tab.component.ts
```diff
@@ -0,0 +1,172 @@
...
+    this.exportsToGCS = store.select('exportToGCS');
+    this.exportsToS3 = store.select('exportToS3');
+
+  }
+
+  ngOnInit() {
+
+    switch (this.source) {
+
+      case 'Down':
```
pshapiro4broad: If you stored the subscriptions in a `Map` then you could replace this switch with something like
```
this.subscription = this.subscriptions.get(this.source).subscribe(...);
```

src/app/file-manager/progress-tab/progress-tab.component.ts
```diff
@@ -0,0 +1,172 @@
...
+          }
+        );
+        break;
+
+    }
+
+  }
+
+  private processChanges(state) {
+    const pending = Object.values(state.pending.items);
```
pshapiro4broad, line 47: Only `inProgress` and `failed` are used below, can the unused ones be removed?

src/app/file-manager/models/item-status.ts
dags: item-status.ts is no longer is active use. It will be removed after a release because has been replaced by EntityStatus, as Transferables related code.

src/app/file-manager/reducers/export-to-gcs.reducer.ts
```diff
@@ -0,0 +1,288 @@
+import { ExportToGCSItem } from '@app/file-manager/models/export-to-gcs-item';
+import * as ExportToGCSItemActions from '@app/file-manager/actions/export-to-gcs-item.actions';
+import { EntityStatus } from '@app/file-manager/models/entity-status';
+
+export interface ExportToGCSState {
```
pshapiro4broad: OK, I see how this will need to be refactored to remove common code.

It might be good to have a comment that provides an overview about what this code is doing, as there's no comments inline.

src/app/file-manager/services/limit-transferables.service.ts
```diff
@@ -1,96 +1,190 @@
...
+  upInProgressCount = 0;
+  upPendingItems: { [id: string]: UploadItem };
+
+  gcsPendingCount = 0;
+  gcsInProgressCount = 0;
+  gcsPendingItems: { [id: string]: ExportToGCSItem };
+
+  s3PendingCount = 0;
+  s3InProgressCount = 0;
+  s3PendingItems: { [id: string]: ExportToS3Item };
```
pshapiro4broad, line 59: This should be modeled using a class that contains the four items for each of the cases. Type generics can be used for the pendingItems field since it will be different for each case.

src/app/file-manager/services/selection.service.ts
```diff
@@ -31,7 +31,9 @@ export class SelectionService {
     }
 
     nothingSelected(): boolean {
-        return this.selection.selected.length <= 0;
+        return this.selection.selected.length === 1
+            && this.selection.selected[0].id === 'workspaces'
+            || this.selection.selected.length <= 0 ? true : false;
```
pshapiro4broad: Why `? true : false`?  Isn't that the same as returning the boolean result of the expression?

src/app/file-manager/transferables-grid/transferables-grid.component.ts
```diff
@@ -1,56 +1,94 @@
...
-  exportToS3Canceled = false;
 
+  // ---------------------------- progress info from here ----------------
+  downCompleted = 0;
+  downTotal = 0;
+  downProgress = 0;
+  downPending = false;
+  downInProgress = false;
+  downCanceled = false;
+  downFinished = false;
```
pshapiro4broad: Using a class for each case would be helpful here as well.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
```diff
@@ -76,97 +192,76 @@ export class TransferablesGridComponent implements OnInit, AfterViewInit {
   applyFilter(filterValue: string) {
     filterValue = filterValue.trim();
     filterValue = filterValue.toLowerCase();
-    this.dataSource.filter = filterValue;
+    this.tabFilter = filterValue;
   }
 
   reset() {
-    this.store.dispatch(new Transferables.Reset());
+    // this.store.dispatch(new Transferables.Reset());
```
pshapiro4broad, line 79: If this code isn't needed, it should be deleted.

pshapiro4broad:
@wloforte 
> That is a mistake, sorry for that. That story is not on review. I moved it back to testing.

OK, make sure it's removed from the PR comment as well.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
dags: yes, we know there are code to refactor here. We'll nedd to re-test after this change, will not be ready until Monday.

src/app/file-manager/transferables-grid/transferables-grid.component.ts
dags: This, and similar functions, are a sort of reminders to features to implement ... It provides some value as documentation and doesn't impact performance or code size. We would like to keep it.

src/app/file-manager/progress-tab/progress-tab.component.ts
dags: Not for now. We want to keep this code until we get feedback from actual users and confirm, or not, the approach implemented.

src/app/file-manager/reducers/export-to-s3.reducer.ts
```diff
@@ -204,23 +204,23 @@ export function ExportToS3Reducer(
             break;
 
         case ExportToS3ItemActions.UPDATE_ITEM_PROGRESS:
-          if (state.inProgress.items[action.payload.id] !== undefined) {
-            // decrement old values
-            state.totalTransferred -= state.inProgress.items[action.payload.id].transferred;
-            state.inProgress.transferred -= state.inProgress.items[action.payload.id].transferred;
+            if (state.inProgress.items[action.payload.id] !== undefined) {
```
pshapiro4broad: The entire file is using the wrong indent; the code uses 2 spaces per indent, not 4.

It looks like all the reducers have the wrong indent setting

main.js
```diff
@@ -68,23 +68,31 @@ app.on('ready', function () {
...
-          { role: 'selectall' }
-        ]
-      }
-    ]));
-  }
+
+    var template = [{
+      label: app.getName(),
+      submenu: [
+        { label: "About " + app.getName(), selector: "orderFrontStandardAboutPanel:" },
```
pshapiro4broad, line 75: This is still incomplete as it doesn't include the Hide menu items. Also the Services menu should be included any time you have selectable text.

Is there a reason why you're not using `role` to define these menu items, e.g.:
```
      {role: 'about'},
      {type: 'separator'},
      {role: 'services', submenu: []},
      {type: 'separator'},
      {role: 'hide'},
      {role: 'hideothers'},
      {role: 'unhide'},
      {type: 'separator'},
      {role: 'quit'}
```

main.js
```diff
@@ -68,23 +68,31 @@ app.on('ready', function () {
...
+        { type: "separator" },
+        { label: "Quit", accelerator: "Command+Q", click: function () { app.quit(); } }
+      ]
+    }, {
+      label: "Edit",
+      submenu: [
+        { label: "Undo", accelerator: "CmdOrCtrl+Z", selector: "undo:" },
```
pshapiro4broad: Using `role` as the code did before seems like a much better way to write this.

src/app/file-manager/progress-tab/progress-tab.component.ts
dags: There is only a subscription per component instance. Code was refactored anyway.

src/app/file-manager/services/limit-transferables.service.ts
dags: I see your point. Could we put in technical debt and create a story to handle it ?. I think it is the only remaining point to fix, right ?

pshapiro4broad: OK to create a tech debt issue. I think that's all the things I found, I'll approve the PR

pshapiro4broad (approved):

